PR #36, opened 2020-05-14
Add custom domain name for api

asjohnston-asf:


.github/workflows/deploy.yml
```diff
@@ -48,4 +49,6 @@ jobs:
               VpcId=${VPC_ID} \
               EDLUsername=${EDL_USERNAME} \
               EDLPassword=${EDL_PASSWORD} \
-              RtcGammaImage=${RTC_GAMMA_REPOSITORY}:${GITHUB_REF##*/}
+              RtcGammaImage=${RTC_GAMMA_REPOSITORY}:${GITHUB_REF##*/} \
+              DomainName=hyp3-${GITHUB_REF##*/}-api.asf.alaska.edu \
```
jhkennedy, line 53: So this would resolve to `hyp3-master-api.asf.alaska.edu` and `hyp3-develop-api.asf.alaska.edu`. I think the develop one is g2g but I'm not terribly a fan of master in that context... maybe we should be using `prod` or `production` instead of `master` as a branch name? But otherwise, this makes sense to me.

PR #39, opened 2020-05-14
drop dead letter queues

asjohnston-asf:
We'll still have cloudwatch metrics to track lambda errors.  DLQs capture the input events for those failures.  Input events for our launch/terminate lambdas are always the same (a fixed launch template/a fixed compute environment), so they're not worth keeping around.


asjohnston-asf:
also lets us drop `sqs:*` permissions from the `cloudfront-deployment-role`

PR #40, opened 2020-05-14
Move Api to Lambda

Jlrine2:


api/SubmitJob/cloudformation.yml
```diff
@@ -0,0 +1,63 @@
+AWSTemplateFormatVersion: 2010-09-09
+
+Outputs:
+
+  LambdaArn:
+    Value: !GetAtt Lambda.Arn
+
+Parameters:
+  JobQueueArn:
```
asjohnston-asf, line 9: blank line

api/SubmitJob/cloudformation.yml
```diff
@@ -0,0 +1,63 @@
...
+    Type: String
+
+  JobDefinition:
```
asjohnston-asf, line 12: this should probably be `JobDefinitionArn` to be specific

api/SubmitJob/cloudformation.yml
```diff
@@ -0,0 +1,63 @@
...
+          - Effect: Allow
+            Action:
+            - logs:CreateLogStream
+            - logs:PutLogEvents
+            Resource: !GetAtt LogGroup.Arn
+          - Effect: Allow
+            Action: batch:SubmitJob
+            Resource:
+              - !Ref JobQueueArn
+              - !Ref JobDefinition
```
asjohnston-asf, line 48: remove an indent level

api/SubmitJob/cloudformation.yml
Jlrine2: I don't think it is the actual Arn, it is the same parameter name we used in the other one

api/SubmitJob/cloudformation.yml
```diff
@@ -0,0 +1,63 @@
...
+
+  Lambda:
+    Type: AWS::Lambda::Function
+    Properties:
+      FunctionName: !Ref AWS::StackName
+      Environment:
+        Variables:
+          JobQueue: !Ref JobQueueArn
+          JobDefinition: !Ref JobDefinition
```
asjohnston-asf, line 58: UPPER_CASE for environment variable names?  don't feel too strongly about it

I also prefer to be specific about whether values are Names or ARNs, `JOB_QUEUE_ARN` and `JOB_DEFINITION_ARN` would be preferable.

I don't think we need it here, but for GRFN I had times where I wanted structured config (which you can't get with just environment variables), so I had a single environment variable with a big json string:

```
Environment:
  Variables:
    Config: !Sub |-
      {
        "job_queue": ${JobQueueArn},
        "job_definition": ${JobDefinition}
      }
```
then `config = json.load(environ['CONFIG'])` or something in the lambda

api/cloudformation.yml
```diff
@@ -15,6 +15,15 @@ Outputs:
...
 
+  SubmitJob:
+    Type: AWS::CloudFormation::Stack
+    Properties:
+      Parameters:
+        JobQueueArn: !Ref JobQueueArn
+        JobDefinition: !Ref JobDefinition
+      TemplateURL: SubmitJob/cloudformation.yml
+
+
```
asjohnston-asf, line 26: remove a blank line here

api/SubmitJob/cloudformation.yml
Jlrine2: I think that config trick could be more usefull if we need to inject more variables, But for how it is now I think this is cleaner.

api/SubmitJob/src/main.py
```diff
@@ -0,0 +1,27 @@
+import json
+from os import getenv
+
+import boto3
+
+client: boto3 = boto3.client('batch')
```
asjohnston-asf, line 6: I don't recognize this synax, I expected `batch = boto3.client('batch')`

api/SubmitJob/cloudformation.yml
asjohnston-asf: Yeah, that's kind of my point, we don't remember what the value is support to look like.  I thinking it's `arn:version`, actually.  Adding `Description`s for all our parameters might be in our future.

api/SubmitJob/src/main.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def lambda_handler(event, context):
+    parameters = json.loads(event['body'])
+    job = client.submit_job(jobName=parameters['granule'],
+                            jobQueue=getenv('JobQueue'),
+                            jobDefinition=getenv('JobDefinition'),
+                            parameters={
```
asjohnston-asf, line 14: this can just be `parameters=parameters`, yes?  lets the function itself stay ignorant of the parameters for a particular processing type?

api/SubmitJob/src/main.py
```diff
@@ -0,0 +1,27 @@
...
+                                'granule': parameters['granule']
+                            }
```
asjohnston-asf, line 16: this can just be `parameters=parameters,`? lets the function itself stay ignorant of the parameters for a particular processing type?

api/SubmitJob/src/main.py
```diff
@@ -0,0 +1,27 @@
...
+                            )
+
+    response = {
+        'Job Name': job['jobName'],
+        'Job Id': job['jobId']
+
```
asjohnston-asf, line 22: unneeded blank line

api/SubmitJob/src/main.py
```diff
@@ -0,0 +1,27 @@
...
+                            jobQueue=getenv('JobQueue'),
+                            jobDefinition=getenv('JobDefinition'),
+                            parameters={
+                                'granule': parameters['granule']
+                            }
+                            )
+
+    response = {
+        'Job Name': job['jobName'],
+        'Job Id': job['jobId']
```
asjohnston-asf, line 21: Not a fan of spaces in dictionary keys.  Can this just be `response=job` to pass along exactly the `submit_job` response? (we'll need to tweak it with more content later, yes)

api/cloudformation.yml
```diff
@@ -48,24 +57,13 @@ Resources:
...
-                       "jobQueue": "${JobQueueArn}",
-                       "parameters": {
-                          "granule" : $input.json('$.granule')
-                       }
-                    }
-                passthroughBehavior: when_no_templates
+                passthroughBehavior: when_no_match
                 httpMethod: POST
-                type: aws
+                type: aws_proxy
```
asjohnston-asf, line 66: could update the order of these elements for easier reading, maybe:
```
type
uri
httpMethod
passthroughBehavior
responses
```

api/SubmitJob/cloudformation.yml
```diff
@@ -0,0 +1,64 @@
...
+      Code: src/
```
asjohnston-asf, line 59: for trivial functions this can point directly at a `.py` file and eliminate the need for the directory, `Code: main.py`, but we'll need a directory back as soon as we need any python packages, e.g. `pip install -r requirements.txt -t src/`

api/SubmitJob/src/main.py
```diff
@@ -0,0 +1,27 @@
+import json
+from os import getenv
```
asjohnston-asf, line 2: I think `environ` is preferable, we'll get a specfic exception when referencing a variable that doesn't exist, rather than `submit_job` complaining about an invalid `None` parameter

api/SubmitJob/src/main.py
asjohnston-asf: or `BATCH = boto3.client('batch')` if we're sticking with PEP8 rules for global names

Jlrine2: That was an artifact of me trying to do a type annotation, turns out boto3 builds all of it's classes based of a json spec, so it wasn't going to work anyway

api/SubmitJob/src/main.py
Jlrine2: The response from batch as a bunch of stuff nobody will care about, it looks like 

```json
{
    "ResponseMetadata": {
        "RequestId": "a435ea3c-d405-4be4-95e5-f18721659387",
        "HTTPStatusCode": 200,
        "HTTPHeaders": {
            "date": "Thu, 14 May 2020 21:25:25 GMT",
            "content-type": "application/json",
            "content-length": "64",
            "connection": "keep-alive",
            "x-amzn-requestid": "a435ea3c-d405-4be4-95e5-f18721659387",
            "x-amz-apigw-id": "MimS6FsoPHcFvsQ=",
            "x-amzn-trace-id": "Root=1-5ebdb745-68db30466536b804ec191c3a;Sampled=0"
        },
        "RetryAttempts": 0
    },
    "jobName": "foo",
    "jobId": "a6481a34-8ef7-4529-90b9-19efa0685e0e"
}
```

api/SubmitJob/cloudformation.yml
Jlrine2: I think I like this because it allows us the ability to easily break the function into multiple files if it starts getting large

api/SubmitJob/cloudformation.yml
Jlrine2: I think I prefer the idea of adding descriptions

api/SubmitJob/src/main.py
asjohnston-asf: think I prefer a minimal response, then:
```
{
  "jobId": job['jobId']
}
```

PR #41, opened 2020-05-14
replace deployment on stack update

Jlrine2:


.github/workflows/deploy.yml
```diff
@@ -35,8 +35,9 @@ jobs:
 
     - name: Deploy to AWS CloudFormation
       run: |
+        sed "s/$REPLACE_ME$/$(openssl rand -hex 4)/" cloudformation.yaml > cloudformation-replaced.yaml
```
asjohnston-asf, line 38: `.yml`, not `.yaml`

asjohnston-asf: not in love with the dollar signs, makes me think they're some special character in the sed command, rather than literal characters to replace

asjohnston-asf: and we want to update `api/cloudformation.yml` rather than the top-level template, which complicates things

asjohnston-asf: since it's a throwaway environment, could do something like `sed -i "s/.../.../" **/cloudformation.yml`

.github/workflows/deploy.yml
```diff
@@ -36,8 +36,9 @@ jobs:
 
     - name: Deploy to AWS CloudFormation
       run: |
+        sed "s/REPLACE_ME/$(openssl rand -hex 4)/" api/cloudformation_unprocessed.yml > api/cloudformation.yml
         aws cloudformation package \
-          --template-file cloudformation.yml \
+          --template-file cloudformation-replaced.yml \
```
jhkennedy, line 41: Should this have the `-replaced`? It looks like you're `sed` is going from `cloudformation_unprocessed.yml` to `cloudforamtion.yml`

.github/workflows/static-analysis.yml
```diff
@@ -47,4 +47,4 @@ jobs:
 
     - name: Run cfn-lint
       # ingore "W3002 This code may only work with `package` cli command" warning
-      run: cfn-lint --ignore-checks W3002 --template *.yml **/*.yml
+      run: cfn-lint --ignore-checks W3002 E3006 --template *.yml **/*.yml
```
jhkennedy, line 50: Why ignore `E3006`?

.github/workflows/deploy.yml
Jlrine2: yes Thanks

.github/workflows/static-analysis.yml
Jlrine2: underscores are invaid in template names, we did it on purpose so that you cant deploy it without doing the replacement

PR #42, opened 2020-05-18
earthdata login implementation for api

asjohnston-asf:


.github/workflows/deploy.yml
```diff
@@ -36,6 +36,7 @@ jobs:
 
     - name: Deploy to AWS CloudFormation
       run: |
+        pip install -r api/submit-job/requirements.txt -t api/submit-job/src/
```
Jlrine2, line 39: would like to see this move to a requirements.txt file

PR #46, opened 2020-05-19
implement authentication

Jlrine2:


api/hyp3_api/auth.py
```diff
@@ -0,0 +1,29 @@
...
+from os import environ
+
+import jwt
+
+
+def decode_token(token, required_scopes):
+    try:
+        jwt.decode(token, environ['AUTH_PUBLIC_KEY'], algorithms=environ['AUTH_ALGORITHM'])
+    except (jwt.DecodeError, jwt.ExpiredSignatureError):
+        return {
```
asjohnston-asf, line 11: formatting could be improved here, `return {'active': True}` would be fine

api/hyp3_api/auth.py
```diff
@@ -0,0 +1,29 @@
+import time
+from os import environ
+
+import jwt
+
+
+def decode_token(token, required_scopes):
+    try:
+        jwt.decode(token, environ['AUTH_PUBLIC_KEY'], algorithms=environ['AUTH_ALGORITHM'])
```
asjohnston-asf, line 9: `decode()` can deduce the algorithm for RSA256, dunno if `environ['AUTH_ALGORITHM']` should be optional or not, probably better to make it required

Jlrine2: It is going to be made required in a future version of PyJWT so is reccomended to be provided

api/hyp3_api/auth.py
```diff
@@ -0,0 +1,29 @@
...
+    except (jwt.DecodeError, jwt.ExpiredSignatureError):
+        return {
+            'active': False
+        }
+    return {
+        'active': True
+    }
+
+
+def get_mock_jwt_cookie(user, expires_in=100):
```
asjohnston-asf, line 19: I like units on variable names where possible, `lifetime_in_seconds` or the like

api/hyp3_api/openapi-spec.yml
```diff
@@ -19,6 +19,9 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/job"
+      security:
+        - URSAuth: []
```
asjohnston-asf, line 23: `earthdata_login` is a better name

api/hyp3_api/openapi-spec.yml
```diff
@@ -46,3 +49,10 @@ components:
       required:
         - granule
       additionalProperties: false
+
+  securitySchemes:
+    URSAuth:
+      type: apiKey
+      name: asf-urs
+      in: cookie
```
asjohnston-asf, line 57: I'd move `in` before `name` so it reads more naturally

api/hyp3_api/auth.py
Jlrine2: Sounds good, I just stole that from Datapool

PR #50, opened 2020-05-22
Connexion

Jlrine2:
This Removes

- Lamda/Apigateway API implementation

This adds:

- Connexion/Flask Api
- Auth integrtation with asf-urs JWT scheme
- modular structure
- api unit tests
- pytest action


api/hyp3_api/__init__.py
```diff
@@ -0,0 +1,8 @@
+import boto3
+import connexion
+
+connexion_app = connexion.App(__name__)
+BATCH_CLIENT = boto3.client('batch')
+
+from hyp3_api import auth, handlers  # noqa Has to be at end of file or will cause circular import
```
asjohnston-asf, line 7: maybe `# noqa (suppress linting, has to be at end of file or will cause circular import)`

PR #55, opened 2020-05-26
Zappa deployment of API

asjohnston-asf:
This is one option.  Using `zappa package` to create the lambda package, then using our own cloudformation to deploy it.  Problems and todos with this approach:

- still not including a custom domain name (*.asf.alaska.edu)
- Not addressing the whole 'update the api gateway deployment with every deploy' issue.  Our gateway is pretty static with this approach, but it'll bite us at some point
- cold start requests take 19 seconds without a zappa keep_warm event
- I dropped all the custom resource names once our stacks got nested deep enough; some resource names are limited to 64 characters (lambda functions, iam roles)
- the lambda handler has to be set to `handler.lambda_handler`
- the stage name in zappa_settings.yml and the AWS::ApiGateway::Stage have to match
- the quickest way to get a dev deployment is becoming "add your feature branch to the deploy.yml action"


Jlrine2:
I think a 19 second cold start time is not good, we should look into fixing that first.

PR #56, opened 2020-05-27
api gateway deployment of api via serverless_wsgi

asjohnston-asf:
I moved `hyp3_api/` down a level to `src/hyp3_api/`, is that the best structure?

still need to fix indentation for lists in all our cloudformation.yml files

still need custom domain name

not sure if we need the `RootMethod` for the api gateway or not

only need CAPABILITY_IAM now instead of CAPABILITY_NAMED_IAM

api gateway stage name is currently `api`, is there anything we like better?  will we hide that prefix when deploying a custom domain name?

are the cloudformation templates reasonably organized?  should we have more, or fewer?  should anything else be moved to the top level template, like the S3 bucket?


.github/workflows/deploy.yml
```diff
@@ -5,11 +5,13 @@ on:
     branches:
     - master
     - develop
+    - asj-serverless
```
asjohnston-asf, line 8: need to revert this before merging

api/src/lambda_handler.py
```diff
@@ -0,0 +1,9 @@
+import serverless_wsgi
+from hyp3_api import connexion_app
+
+
+serverless_wsgi.TEXT_MIME_TYPES.append('application/problem+json')
+
+
+def handler(event, context):
+    return serverless_wsgi.handle_request(connexion_app.app, event, context)
```
asjohnston-asf, line 9: should we write a test for this?  at least invoke with a typical api gateway event and check that we get a 404 or 401 resopnse back?

api/cloudformation.yml
```diff
@@ -0,0 +1,144 @@
...
+
+  Url:
+    Value: !Sub "https://${RestApi}.execute-api.${AWS::Region}.amazonaws.com/${Stage}/"
+
+Resources:
+
+  RestApi:
+    Type: AWS::ApiGateway::RestApi
+    Properties:
+      Name: !Ref AWS::StackName
```
asjohnston-asf, line 26: we could describe the entire api using the `body` attribute and an openapi spec instead of the separate `Resource` and `Method` items, not sure which I like

api/src/lambda_handler.py
Jlrine2: I don't know that this really needs a test, it's just a call to a library function.

PR #58, opened 2020-05-27
production api gateway deployment

asjohnston-asf:
- deploy api via api gateway and `serverless_wsgi`
- remove `jobName` field from /jobs json response
- remove custom aws resource names


.github/workflows/deploy.yml
```diff
@@ -31,18 +33,28 @@ jobs:
...
-          --template-file batch-cluster/cloudformation.yml \
+          --template-file cloudformation.yml \
           --s3-bucket ${TEMPLATE_BUCKET} \
           --output-template-file packaged.yml
         aws cloudformation deploy \
           --stack-name hyp3-${GITHUB_REF##*/} \
           --template-file packaged.yml \
           --role-arn ${CLOUDFORMATION_ROLE_ARN} \
-          --capabilities CAPABILITY_NAMED_IAM \
+          --capabilities CAPABILITY_IAM \
```
jhkennedy, line 54: Should `CAPABILITY_IAM` be a variable? I'm not sure where it's coming from

Jlrine2: It's a flag that is required to create IAM roles, CAPABILITY_NAMED_IAM is a flag that is required to create IAM roles with custom names

asjohnston-asf: https://docs.aws.amazon.com/cli/latest/reference/cloudformation/deploy/index.html
```
--capabilities (list) A list of capabilities that you must specify before AWS Cloudformation can create certain stacks. Some stack templates might include resources that can affect permissions in your AWS account, for example, by creating new AWS Identity and Access Management (IAM) users. For those stacks, you must explicitly acknowledge their capabilities by specifying this parameter. The only valid values are CAPABILITY_IAM and CAPABILITY_NAMED_IAM. If you have IAM resources, you can specify either capability. If you have IAM resources with custom names, you must specify CAPABILITY_NAMED_IAM. If you don't specify this parameter, this action returns an InsufficientCapabilities error.
```

asjohnston-asf: It's effectively the checkbox in the web console that says "I acknowledge this stack might create IAM resources"

jhkennedy: Ah great, it just looked variable-y being in all caps. Thanks for the docs.

PR #60, opened 2020-05-28
add custom domain name for api

asjohnston-asf:
This will make the two apis available at
https://hyp3-test-api.asf.alaska.edu/ui
https://hyp3-api.asf.alaska.edu/ui

This makes it impossible to deploy the api stack without providing a valid certificate arn.  That could be too restrictive.  We could move the DomainName and BasePathMapping resources out of the api stack and into the top level hyp3/cloudformation.yml, but I don't know that that buys us much.

We could also reject this PR, and make configuring the custom domain name a completely manual process, but I don't like that approach much, either.  I'm comfortable moving forward as-is for now.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -60,5 +60,5 @@ jobs:
                 "EDLPassword=${EDL_PASSWORD}" \
                 "RtcGammaImage=${RTC_GAMMA_REPOSITORY}:${GITHUB_REF##*/}" \
                 "AuthPublicKey=${AUTH_PUBLIC_KEY}" \
-                "DomainName=hyp3-${GITHUB_REF##*/}.asf.alaska.edu" \
+                "DomainName=hyp3-${GITHUB_REF##*/}-api.asf.alaska.edu" \
                 "CertificateArn=${CERTIFICATE_ARN}"
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - master
       - develop
+      - asj-domain-name
 
 env:
   AWS_REGION: us-west-2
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -57,4 +58,6 @@ jobs:
                 "EDLUsername=${EDL_USERNAME}" \
                 "EDLPassword=${EDL_PASSWORD}" \
                 "RtcGammaImage=${RTC_GAMMA_REPOSITORY}:${GITHUB_REF##*/}" \
-                "AuthPublicKey=${AUTH_PUBLIC_KEY}"
+                "AuthPublicKey=${AUTH_PUBLIC_KEY}" \
+                "DomainName=hyp3-${GITHUB_REF##*/}-api.asf.alaska.edu" \
```
Jlrine2, line 62: I'm not a huge fan of this, the prod api should just be hype-api not hype-master-api

api/cloudformation.yml
```diff
@@ -13,17 +13,26 @@ Parameters:
...
+    Type: String
+
+  CertificateArn:
+    Type: String
+
 Outputs:
 
   Url:
-    Value: !Sub "https://${RestApi}.execute-api.${AWS::Region}.amazonaws.com/${Stage}/"
+    Value: !Sub "https://${DomainName}/ui"
```
Jlrine2, line 25: Do we want to return a reference to the UI endpoint, or root?

asjohnston-asf: UI was the one I wanted all the time, so I thought I'd save myself the work.

Now that we'll have static domain names for the two deployments this output parameter will probably only be leveraged by ad-hoc deployments, in which case I'll want to straight to the UI for testing I think.

asjohnston-asf: although ad-hoc deployments probably won't leverage, `https://${DomainName}`, they'll go straight to the gateway at `https://${RestApi}.execute-api.${AWS::Region}.amazonaws.com/${Stage}/`

PR #63, opened 2020-05-29
Reorganize post schema

Jlrine2:
in order to capture email, and in the future other user and job information, we need a bit more verbose of an post schema

### DIFF
--- a/api/src/hyp3_api/handlers.py
+++ b/api/src/hyp3_api/handlers.py
@@ -6,7 +6,6 @@ from hyp3_api import BATCH_CLIENT, connexion_app
 
 def submit_job(body):
     parameters = body['parameters']
-    print(parameters)
     job = BATCH_CLIENT.submit_job(
         jobName=parameters['granule'],
         jobQueue=environ['JOB_QUEUE'],
--- a/api/src/hyp3_api/handlers.py
+++ b/api/src/hyp3_api/handlers.py
@@ -5,16 +5,20 @@ from hyp3_api import BATCH_CLIENT, connexion_app
 
 
 def submit_job(body):
+    parameters = body['parameters']
+    print(parameters)
     job = BATCH_CLIENT.submit_job(
-        jobName=body['granule'],
+        jobName=parameters['granule'],
         jobQueue=environ['JOB_QUEUE'],
         jobDefinition=environ['JOB_DEFINITION'],
-        parameters=body
+        parameters=parameters
     )
+    print(job)
     response = {
         'jobId': job['jobId'],
-        'parameters': body,
+        'parameters': parameters,
     }
+
     return response
 
 
--- a/api/src/hyp3_api/openapi-spec.yml
+++ b/api/src/hyp3_api/openapi-spec.yml
@@ -47,6 +47,7 @@ components:
             - RTC_GAMMA
         email:
           type: string
+          format: email
           example: john.doe@example.com
         parameters:
           $ref: "#/components/schemas/parameters"
--- a/api/src/hyp3_api/openapi-spec.yml
+++ b/api/src/hyp3_api/openapi-spec.yml
@@ -12,7 +12,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/parameters"
+              $ref: "#/components/schemas/submission"
         required: true
       responses:
         "200":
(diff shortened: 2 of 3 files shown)


api/src/hyp3_api/openapi-spec.yml
```diff
@@ -37,6 +37,22 @@ components:
...
+          type: string
+          example: RTC_GAMMA
+        email:
+          type: string
+          example: john.doe@example.com
+        parameters:
+          $ref: "#/components/schemas/parameters"
+      required:
+        - process_type
+        - email
```
Jlrine2, line 53: not actually implemented yet, thats next on my list

api/src/hyp3_api/openapi-spec.yml
```diff
@@ -37,6 +37,22 @@ components:
...
+        process_type:
+          type: string
+          example: RTC_GAMMA
+        email:
+          type: string
+          example: john.doe@example.com
+        parameters:
+          $ref: "#/components/schemas/parameters"
+      required:
+        - process_type
```
Jlrine2, line 52: Don't know that we need this until we support multiple process_types

api/src/hyp3_api/openapi-spec.yml
```diff
@@ -37,6 +37,22 @@ components:
...
       additionalProperties: false
+    submission:
+      type: object
+      properties:
+        process_type:
+          type: string
+          example: RTC_GAMMA
+        email:
+          type: string
+          example: john.doe@example.com
```
Jlrine2, line 48: considered a pattern, not sure if worth it as emails can actually be pretty tricky to match well

asjohnston-asf: `type: email` is a build-in jsonschema type
https://json-schema.org/understanding-json-schema/reference/string.html#email-addresses

api/src/hyp3_api/openapi-spec.yml
```diff
@@ -37,6 +37,22 @@ components:
         - jobId
         - parameters
       additionalProperties: false
+    submission:
+      type: object
+      properties:
+        process_type:
+          type: string
+          example: RTC_GAMMA
```
asjohnston-asf, line 45: can use `enum` to restrict the possible choices here
```
enum:
  - RTC_GAMMA
```

api/src/hyp3_api/openapi-spec.yml
Jlrine2: Makes our test fail... Sad day

asjohnston-asf: https://swagger.io/specification/#data-type-format

looks like this might work instead:
```
email:
  type: string
  format: email
  example: john.doe@example.com
```

api/src/hyp3_api/handlers.py
```diff
@@ -5,16 +5,19 @@
 
 
 def submit_job(body):
+    parameters = body['parameters']
```
asjohnston-asf, line 8: might `print(body)` here, we don't get any visibility into the body from access logs

api/src/hyp3_api/handlers.py
```diff
@@ -5,16 +5,19 @@
...
         jobDefinition=environ['JOB_DEFINITION'],
-        parameters=body
+        parameters=parameters
     )
+    print(job)
     response = {
         'jobId': job['jobId'],
-        'parameters': body,
+        'parameters': parameters,
     }
```
asjohnston-asf, line 19: `response` could include everything they gave us, plus the job id?  so `body[jobId] = job['jobId']`; return job`?

api/tests/test_hyp3_api.py
```diff
@@ -23,10 +23,17 @@ def batch_stub():
...
         add_response(batch_stub, granule)
-    return client.post(JOBS_URI, json={'granule': granule})
+    payload = {
+        'process_type': 'RTC_GAMMA',
+        'email': email,
+        'parameters': {
+            'granule': granule
+        }
+    }
+    return client.post(JOBS_URI, json=payload)
```
asjohnston-asf, line 38: missing email, invalid email, missing process type, invalid process type, on the fence about testing these but I think once I've bothered to write tests I'm almost always happy to have them

PR #72, opened 2020-06-04
Update jobs table from step function

asjohnston-asf:


Jlrine2:
This adds:
- Lamda update function, job_id is required key, updates all other fields, creates if no record with that job_id
- Job_running state added to state machine, calls Lamda update DB with JobID and execution Time
_ Failed and succeded states

PR #84, opened 2020-06-04
add optional description on job submission

asjohnston-asf:


api/src/hyp3_api/openapi-spec.yml
```diff
@@ -79,6 +81,9 @@ components:
       enum:
         - RTC_GAMMA
 
+    description:
+      type: string
```
asjohnston-asf, line 85: tempted to add `minLength: 1` here

PR #92, opened 2020-06-09
APD-1523: add s3 access logging for content bucket

asjohnston-asf:
https://docs.aws.amazon.com/AmazonS3/latest/dev/ServerLogs.html

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-loggingconfig.html

### DIFF
--- a/cloudformation.yml
+++ b/cloudformation.yml
@@ -77,9 +77,22 @@ Resources:
         JobsTable: !Ref JobsTable
       TemplateURL: step-function/cloudformation.yml
 
+  LogBucket:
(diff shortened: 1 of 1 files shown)


asjohnston-asf:
Longer term, once we know how logs are getting used in practice, we might consider a bucket policy to deny s3:DeleteObject on the log bucket.  If/when we have a lot of logs we might also consider a lifecycle policy to delete old logs, or at least aggregate them into a more manageable format.

PR #110, opened 2020-06-12
fix user data script for amazon-linux-2 ecs ami

asjohnston-asf:
This just does a straight `mount` command, rather than putting a permanent entry in `/etc/fstab`.  If anyone knows the details of how that works and wants to do it that way, go for it.

I'm not confident if `cloud-init-per instance` is the right frequency (`once` is the other option), or if we should even bother with `cloud-init-per` at all.


jhkennedy:
from my brief poking around, `per` seems to be the way to go

PR #118, opened 2020-06-16
start/end parameters for GET /jobs

asjohnston-asf:
The dynamodb filter is just doing string comparison.  I think that's compatible with date-time, unless `format: date-time` allows for time zones other than `Z`.

Need `strict-rfc3339` in requirements.txt for `format: date-time`, see https://github.com/zalando/connexion/issues/476

Enabled `strict_validation=True` so we now throw HTTP 400 if the user provides query parameters not listed in the spec.

I'm starting to like `request_time` better than `start_time` for the field name in the DB table.  I'd also consider `before` and `after` for the query parameters rather than `start` and `end`; I think it makes the comparisons more self-evident.

### DIFF
--- a/api/src/hyp3_api/handlers.py
+++ b/api/src/hyp3_api/handlers.py
@@ -37,9 +37,9 @@ def get_jobs(user, status_code=None, start=None, end=None):
     if status_code is not None:
         filter_expression = filter_expression & Attr('status_code').eq(status_code)
     if start:
-        filter_expression = filter_expression & Attr('start_time').gte(start)
+        filter_expression = filter_expression & Attr('request_time').gte(start)
     if end:
-        filter_expression = filter_expression & Attr('start_time').lte(end)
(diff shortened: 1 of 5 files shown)


api/src/hyp3_api/handlers.py
```diff
@@ -31,11 +31,15 @@ def post_jobs(body, user):
     return [{'job_id': job_id} for job_id in job_ids]
 
 
-def get_jobs(user, status_code=None):
+def get_jobs(user, status_code=None, start=None, end=None):
     table = DYNAMODB_RESOURCE.Table(environ['TABLE_NAME'])
     filter_expression = Attr('job_id').exists()
     if status_code is not None:
```
Jlrine2, line 37: this could use some form of refactor

jhkennedy:
I agree with `request_time` over `start_time` -- seems to be more in line with what we're actually recording, especially so with retries

PR #128, opened 2020-06-17
refactor support for decimal datatypes

asjohnston-asf:
https://stackoverflow.com/questions/16957275/python-to-json-serialization-fails-on-decimal

https://stackoverflow.com/questions/19965018/python-decimal-checking-if-integer


Jlrine2:
I think this can be closed
https://github.com/asfadmin/hyp3/pull/129

PR #138, opened 2020-06-22
Implement user quotas

Jlrine2:


api/src/hyp3_api/handlers.py
```diff
@@ -24,7 +25,9 @@ def post_jobs(body, user):
     print(body)
     if not context['is_authorized']:
         abort(403)
-
+    job_count = get_job_count_for_month(user)
+    if job_count + len(body['jobs']) > int(environ['MONTHLY_JOB_QUOTA_PER_USER']):
+        abort(400)
```
asjohnston-asf, line 30: per https://connexion.readthedocs.io/en/latest/response.html#returning-status-codes, this might just be:
`return f'....', 400`

PR #139, opened 2020-06-23
add meaningful 403 forbidden error message

asjohnston-asf:


api/src/hyp3_api/handlers.py
```diff
@@ -28,7 +27,7 @@ class QuotaError(Exception):
 def post_jobs(body, user):
     print(body)
     if not context['is_authorized']:
-        abort(403)
+        return problem(403, 'Forbidden', 'You do not have permission to submit jobs.')
```
asjohnston-asf, line 30: maybe `User  {user} does not have permission to submit jobs.` would be better?  I imagine mixing up EDL usernames will be somewhat common, at least internally at ASF.

jhkennedy: I agree, putting the username in there would be best.

api/src/hyp3_api/handlers.py
```diff
@@ -28,7 +27,7 @@ class QuotaError(Exception):
 def post_jobs(body, user):
     print(body)
     if not context['is_authorized']:
-        abort(403)
+        return problem(403, 'Forbidden', f'User {user} does not have permission to submit jobs.')
```
Jlrine2, line 30: This might be an okay place for flask status module

PR #140, opened 2020-06-23
Add db index for status_code

asjohnston-asf:


api/tests/test_hyp3_api.py
```diff
@@ -23,48 +24,23 @@ def client():
...
-                'WriteCapacityUnits': 15,
-            },
+            **table_properties,
         )
         yield table
 
 
+def get_table_properties_from_template():
+    yaml.SafeLoader.add_multi_constructor('!', lambda loader, suffix, node: None)
+    with open('../cloudformation.yml', 'r') as f:
```
asjohnston-asf, line 38: This is from the working directory, which with pytest could be anywhere.  If we want to pursue this approach we'll need to make it relative to the `test_hyp3_api.py` file's directory, or load it in explicitly as an environment variable.

Jlrine2: I think os.path has a tool for this?

PR #146, opened 2020-06-26
move tests to more specific files for testing

Jlrine2:


api/tests/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+def get_table_properties_from_template():
+    yaml.SafeLoader.add_multi_constructor('!', lambda loader, suffix, node: None)
+    template_file = path.join(path.dirname(__file__), '../../cloudformation.yml')
+    with open(template_file, 'r') as f:
+        template = yaml.safe_load(f)
+    table_properties = template['Resources']['JobsTable']['Properties']
+    return table_properties
+
+
+def make_job(granule='S1B_IW_SLC__1SDV_20200604T082207_20200604T082234_021881_029874_5E38',
```
asjohnston-asf, line 44: `make_job` and `submit_batch` are specific to `POST /jobs`, I was thinking we'd implement them in `test_submit_jobs.py`

Jlrine2: similarly we could take `make_db_record` into `test_list_jobs.py`

api/tests/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+        'job_parameters': {
+            'granule': granule
+        }
+    }
+    if description is not None:
+        job['description'] = description
+    return job
+
+
+def make_db_record(job_id,
```
asjohnston-asf, line 58: only used by `test_list_jobs.py`, might move the definition there

api/tests/test_api_spec.py
```diff
@@ -0,0 +1,156 @@
+from conftest import AUTH_COOKIE, DEFAULT_USERNAME, JOBS_URI, auth, login, make_job, submit_batch
+from flask_api import status
+
+
+def test_not_logged_in(client):
+    response = submit_batch(client)
```
asjohnston-asf, line 6: think this can be `response = client.post(JOBS_URI)`, actually, it should check auth before checking the payload

api/tests/conftest.py
Jlrine2: `test_submit_job` and `submit_batch` are used in `test_api_spec.py`

api/tests/test_api_spec.py
```diff
@@ -0,0 +1,156 @@
...
+    assert response.status_code == status.HTTP_401_UNAUTHORIZED
+
+    response = client.get(JOBS_URI)
+    assert response.status_code == status.HTTP_401_UNAUTHORIZED
+
+    response = client.head(JOBS_URI)
+    assert response.status_code == status.HTTP_401_UNAUTHORIZED
+
+
+def test_logged_in_not_authorized(client):
```
asjohnston-asf, line 16: authorization only applies to `POST /jobs`, might move this test to `test_submit_jobs.py`, along with the granule name tests

api/tests/conftest.py
asjohnston-asf: hrmm, I guess the quota check in `test_submit_jobs.py` needs it, so might as well leave it here

api/tests/test_api_spec.py
Jlrine2: I was going for a model where the things that get to the handlers are in the respective files, things checked by the api are in `test_api_spec.py`. Auth is kinda weird because the logic is done before the handler, but sets a variable checked at the handler

api/tests/test_api_spec.py
```diff
@@ -0,0 +1,156 @@
+from conftest import AUTH_COOKIE, DEFAULT_USERNAME, JOBS_URI, auth, login, make_job, submit_batch
```
asjohnston-asf, line 1: importing `auth` from `conftest` threw me, think we should import it from `hyp3_api` where it's defined

api/tests/test_api_spec.py
asjohnston-asf: `test_not_logged_in()` and `test_jobs_bad_method()` could both reduce to either of these, I kind of like the `open()` approach:

```
    for method in [client.get, client.post, client.head]:
        response = method(JOBS_URI)
        assert response.status_code == status.HTTP_401_UNAUTHORIZED

    for method in ['POST', 'GET', 'HEAD']:
        response = client.open(JOBS_URI, method=method)
        assert response.status_code == status.HTTP_401_UNAUTHORIZED
```

api/tests/test_api_spec.py
```diff
@@ -0,0 +1,156 @@
...
+    response = client.post(JOBS_URI, headers={'Origin': 'https://www.google.com'})
+    assert 'Access-Control-Allow-Origin' not in response.headers
+    assert 'Access-Control-Allow-Credentials' not in response.headers
+
+    response = client.post(JOBS_URI, headers={'Origin': 'https://www.alaska.edu'})
+    assert 'Access-Control-Allow-Origin' not in response.headers
+    assert 'Access-Control-Allow-Credentials' not in response.headers
+
+
+def test_cors_good_origins(client):
```
asjohnston-asf, line 145: this deserves a loop, too:
```
for origin in ['.....']:
    response = client.post(JOBS_URI, headers={'Origin': origin})
    assert response.headers['Access-Control-Allow-Origin'] == origin
    assert response.headers['Access-Control-Allow-Credentials'] == 'true'
```

api/tests/test_api_spec.py
```diff
@@ -0,0 +1,156 @@
...
+    assert response.status_code == status.HTTP_404_NOT_FOUND
+
+
+def test_cors_no_origin(client):
+    response = client.post(JOBS_URI)
+    assert 'Access-Control-Allow-Origin' not in response.headers
+    assert 'Access-Control-Allow-Credentials' not in response.headers
+
+
+def test_cors_bad_origins(client):
```
asjohnston-asf, line 135:
```python
for origin in ['.....']:
    response = client.post(JOBS_URI, headers={'Origin': origin})
    assert 'Access-Control-Allow-Origin' not in response.headers
    assert 'Access-Control-Allow-Credentials' not in response.headers
```

api/tests/test_api_spec.py
asjohnston-asf: There's still some spec validation in `test_submit_jobs.py`, e.g. too many jobs, missing description, etc.  I feel more strongly about moving the granule name checks than moving the auth check, but moving both puts everything that needs to know about the structure of a new job payload into the one `test_submit_jobs.py` file.

api/tests/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+def client():
+    with connexion_app.app.test_client() as test_client:
+        yield test_client
+
+
+@pytest.fixture()
+def table():
+    table_properties = get_table_properties_from_template()
+    with mock_dynamodb2():
+        boto3.setup_default_session()
```
asjohnston-asf, line 27: I think this can go, along with `import boto3`? docs suggest calling it without parameters is the default behavior anyway

PR #148, opened 2020-06-26
convert time to isoformat 8601 for human readability

Jlrine2:


api/src/hyp3_api/handlers.py
```diff
@@ -1,12 +1,13 @@
 from datetime import datetime
 from decimal import Decimal
 from os import environ
-from time import time
 from uuid import uuid4
 
+import pytz
```
Jlrine2, line 6: I'm a bit tempted to define UTC ourselves so we can skip this import. 
```python
class utc_tz(tz_info):
    def utcoffset(self, dt):
        return timedelta(0)
    def tzname(self, dt):
        return '+00:00'

File: api/src/hyp3_api/openapi-spec.yml [hunk 2]
```diff
@@ -156,9 +156,8 @@ components:
       description: The name of the Sentinel-1 granule to process.  Only SLC products with beam mode IW are supported.
 
     datetime:
-      type: integer
-      minimum: 0
-      description: a date time in unix epoch time
+      type: string
```
Jlrine2, line 159: would love to add a custom validator here, date-time isn't working for me

asjohnston-asf: `date-time` needs `pip install strict-rfc3339`

Jlrine2: I have it and it wasn't working

api/src/hyp3_api/handlers.py
```diff
@@ -34,7 +35,7 @@ def post_jobs(body, user):
     except QuotaError as e:
         return problem(400, 'Bad Request', str(e))
 
-    request_time = int(time())
+    request_time = datetime.utcnow().isoformat('T') + 'Z'
```
asjohnston-asf, line 38: https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat

`sep='T'` is the default, don't know that we need to call it out.  I also think we should set `timespec='seconds'`; milisecond precision just seems like a distraction.

`isoformat` gets calls three places in handlers and twice in tests, think it deserves a helper function:

```python
def format_datetime(value: datetime):
    utc_datetime = convert_to_utc(value)  # or implement inline here
    formatted_datetime = utc_datetime.isoformat(timespec='seconds') + 'Z'
    return formatted_datetime
```

PR #150, opened 2020-06-30
Release: query by start date, end date, readable date formats

asjohnston-asf:
**This deployment will delete all records in the production dynamodb table.**  As of this writing the prod db only has two records, both for `asjohnston`.  The corresponding S3 objects will not be automatically deleted, but our lifecycle rules will remove them within two weeks.

Changed:
- `request_time` and `expiration_date` are now formatted `YYYY-MM-DDTHH24:MI:SS+00:00` instead of unix epoch time in `GET /jobs` and `PUT/ jobs` responses

Added:
- new `start` and `end` query parameters for `GET /jobs` allowing users to filter results to jobs with `request_time` between the given dates

Technical changes:
- segregated api tests into multiple files
- datetime fields are now strings formatted `YYYY-MM-DDTHH24:MI:SS+00:00` in dynamodb (instead of unix timestamp integers)


asjohnston-asf:
holding off until we have the timezone handling finalized in develop

PR #151, opened 2020-06-30
add end query parameter to GET /jobs

asjohnston-asf:
Also:
- format datetimes in database with `+00:00` time zone instead of `Z`
- make all datetime objects timezone-aware
- use `datetime.timezone.utc` instead of `pytz.UTC`

### DIFF
--- a/step-function/get-files/src/main.py
+++ b/step-function/get-files/src/main.py
@@ -17,7 +17,7 @@ def get_expiration_time(key):
     s3_object = S3_CLIENT.get_object(Bucket=environ['BUCKET'], Key=key)
     expiration_string = s3_object['Expiration'].split('"')[1]
     expiration_datetime = datetime.strptime(expiration_string, '%a, %d %b %Y %H:%M:%S %Z')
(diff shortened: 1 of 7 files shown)


step-function/get-files/src/main.py
```diff
@@ -17,7 +17,7 @@ def get_expiration_time(key):
     s3_object = S3_CLIENT.get_object(Bucket=environ['BUCKET'], Key=key)
     expiration_string = s3_object['Expiration'].split('"')[1]
     expiration_datetime = datetime.strptime(expiration_string, '%a, %d %b %Y %H:%M:%S %Z')
```
asjohnston-asf, line 19: Haven't found a way to get a timezone-aware datetime out of this.  `expiration_string` is of the form `Fri, 10 Jul 2020 00:00:00 GMT`.  `GMT` is a timezone name, (%Z), not a timezone offset (%z).  See notes for %z/%Z at https://docs.python.org/3/library/datetime.html#technical-detail

PR #152, opened 2020-07-07
APD: 1560 Check granule exists in ASF Archive

Jlrine2:


api/tests/conftest.py
```diff
@@ -84,5 +89,16 @@ def make_db_record(job_id,
...
 
 
+def setup_requests_mock(batch):
+    granules = [job['job_parameters']['granule'] for job in batch]
+    cmr_response = {
+        'feed':
+            {
+                'entry': [{'producer_granule_id': granule} for granule in granules]
+            }
+    }
```
jhkennedy, line 99: Only thing I see, and is rather nitpick-y, is I'd probably write this like:
```py
     cmr_response = {
        'feed': {
            'entry': [{'producer_granule_id': granule} for granule in granules]
        }
    }
```

api/requirements.txt
```diff
@@ -1,11 +1,12 @@
...
+connexion[swagger-ui]
+cryptography
 flask_api
+Flask-Cors
+moto
 PyJWT
-cryptography
+pytest
+PyYAML
+responses
```
asjohnston-asf, line 10: need to add

api/src/hyp3_api/handlers.py
```diff
@@ -34,6 +39,13 @@ def post_jobs(body, user):
     except QuotaError as e:
         return problem(400, 'Bad Request', str(e))
 
+    try:
+        check_granules_exists([job['job_parameters']['granule'] for job in body['jobs']])
```
asjohnston-asf, line 43: could include ` if job['job_type'] == 'RTC_GAMMA'`, not strictly necessary yet though

asjohnston-asf: and maybe break up the creation of the list and the function call:
```
granules = ....
check_granules_exist(granules)
```

api/src/hyp3_api/handlers.py
```diff
@@ -102,6 +114,23 @@ def get_request_time_expression(start, end):
         return key.lte(formatted_end)
 
 
+def check_granules_exists(granules):
```
asjohnston-asf, line 117: `check_granules_exist` without the `s` at the end

api/src/hyp3_api/handlers.py
```diff
@@ -102,6 +114,23 @@ def get_request_time_expression(start, end):
...
+        'provider': 'ASF',
+        'collection_concept_ids': {
+            "C1214470488-ASF",  # SENTINEL-1A_SLC
+            "C1327985661-ASF",  # SENTINEL-1B_SLC
+        }
+    }
+    response = requests.get('https://cmr.earthdata.nasa.gov/search/granules.json', params=cmr_parameters)
+    response.raise_for_status()
+    found_granules = [entry['producer_granule_id'] for entry in response.json()['feed']['entry']]
+    if set(granules) != set(found_granules):
```
asjohnston-asf, line 129: you can use the subtraction operator on sets to the the difference, https://docs.python.org/2/library/sets.html#set-objects
```
not_found_granules = granules - found_granules
if not_found_granules:
    raise CmrError(f'Requested scenes could not be found: {",".join(not_found_granules)}')
```

api/tests/conftest.py
```diff
@@ -12,6 +15,8 @@
 DEFAULT_JOB_ID = 'myJobId'
 DEFAULT_USERNAME = 'test_username'
 
+CMR_URL_RE = re.compile('https://cmr.earthdata.nasa.gov/search/granules.json.*')
```
asjohnston-asf, line 18: might put `CMR_URL = 'https://cmr.earthdata.nasa.gov/search/granules.json'` in `__init__.py` then import it in both `handlers.py` and `conftest.py`

api/tests/test_submit_job.py
```diff
@@ -70,5 +77,25 @@ def test_submit_job_with_empty_description(client):
...
+
+    login(client)
+    response = submit_batch(client, batch)
+    assert response.status_code == status.HTTP_200_OK
+
+    response = submit_batch(client, [make_job('S1A_IW_SLC__1SDV_20200610T173646_20200610T173704_032958_03D14C_5F2A')])
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
+    assert response.json['title'] == 'Bad Request'
+    assert response.json['detail'] == 'Requested scenes could not be found: ' \
+                                      'S1A_IW_SLC__1SDV_20200610T173646_20200610T173704_032958_03D14C_5F2A'
```
asjohnston-asf, line 101: Think I'd combine these into a single request with some good granules and some bad granules to verify the error message only complains about the bad ones.  The `200` test doesn't add anything on its own over other test cases for submitting jobs.

api/tests/test_submit_job.py
```diff
@@ -70,5 +77,25 @@ def test_submit_job_with_empty_description(client):
     batch = [
         make_job(description='')
     ]
+    setup_requests_mock(batch)
     response = submit_batch(client, batch)
     assert response.status_code == status.HTTP_400_BAD_REQUEST
+
+
+@responses.activate
```
asjohnston-asf, line 85: why is this the only test case with a decorator?

asjohnston-asf:
`handlers.py` is starting to feel a little bloated, might move the helper functions elsewhere and just leave the endpoint functions (similar to what we did with flask views.py)

api/tests/test_submit_job.py
```diff
@@ -70,5 +77,25 @@ def test_submit_job_with_empty_description(client):
...
+    setup_requests_mock(batch)
+
+    login(client)
+    response = submit_batch(client, batch)
+    assert response.status_code == status.HTTP_200_OK
+
+    response = submit_batch(client, [make_job('S1A_IW_SLC__1SDV_20200610T173646_20200610T173704_032958_03D14C_5F2A')])
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
+    assert response.json['title'] == 'Bad Request'
+    assert response.json['detail'] == 'Requested scenes could not be found: ' \
```
asjohnston-asf, line 100: `Some requested scenes...` ?

PR #156, opened 2020-07-08
added changelog and relaese workflow

Jlrine2:


.github/workflows/deploy.yml
```diff
@@ -74,3 +77,19 @@ jobs:
...
+        run: |
+          git fetch origin +refs/tags/*:refs/tags/*
+          git config user.email "UAF-asf-apd@alaska.edu"
+          git config user.name "tools-bot"
+          LABEL_QUERY='.data.repository.pullRequest.labels.nodes[].name'
+          SELECT='select(. == "major" or . == "minor" or . == "patch")'
+          BUMP_PART=$(jq --raw-output  "${LABEL_QUERY} | ${SELECT}" labels.json | sort | head -1)
+          PR_QUERY='.data.repository.commit.associatedPullRequests.edges[0].node.title'
+          TAG_MSG=$(jq --raw-output "${PR_QUERY}"  pr.json)
+          bump2version --current-version $(git describe --abbrev=0) \
```
jhkennedy, line 92: I think you need to add bump2version to `api/requirements.txt` or, pip install it after

https://github.com/asfadmin/hyp3/blob/476e622089a27fd6f71bdd3df3169c42393a204d/.github/workflows/deploy.yml#L60

jhkennedy: yeah, just install it after so it's not in src

PR #170, opened 2020-07-21
add description fields to the api-spec

Jlrine2:


api/src/hyp3_api/openapi-spec.yml
```diff
@@ -51,6 +53,7 @@ paths:
                 $ref: "#/components/schemas/jobs_response"
   /user:
     get:
+      description: Get information about a user.
```
asjohnston-asf, line 56: about the logged in user

api/src/hyp3_api/openapi-spec.yml
```diff
@@ -169,21 +176,25 @@ components:
           $ref: "#/components/schemas/datetime"
 
     job_id:
+      description: Unique identifier for a job
       type: string
       format: uuid
       example: 27836b79-e5b2-4d8f-932f-659724ea02c3
 
     user_id:
+      description: Username from EarthData Login.
```
asjohnston-asf, line 185: Earthdata

PR #178, opened 2020-07-24
add support for grdh iw granules

asjohnston-asf:


api/tests/test_submit_job.py
```diff
@@ -99,3 +99,46 @@ def test_submit_job_granule_does_not_exist(client, table):
...
+        response = submit_batch(client, batch)
+        assert response.status_code == status.HTTP_200_OK
+
+
+def test_submit_bad_granule_names(client):
+    login(client)
+    bad_granule_names = [
+        'foo',
+        'S1B_IW_SLC__1SDV_20200604T082207_20200604T082234_021881_029874_5E3',
+        'S1B_IW_SLC__1SDV_20200604T082207_20200604T082234_021881_029874_5E38_',
```
jhkennedy, line 126: Might want an `S1C` bad granule as well -- nothing is checking the first part of the regex currently

PR #181, opened 2020-07-24
Create a method for shutting down HyP3

Jlrine2:
Adds a parameter to the root CF stack, set to `false` to shut down the ability to submit jobs.


cloudformation.yml
```diff
@@ -48,6 +48,14 @@ Parameters:
     MinValue: 1
     Default: 25
 
+  SystemAvailable:
+    Description: Set to false to shutdown system, API will run and provide errors to users, but will not accept jobs.
```
Jlrine2, line 52: I wonder if this should be called out specifically in the readme, or if the note to review parameters is sufficient.

If I wasn't familiar with this app but needed to shut it down, this might be hard to find.

api/src/hyp3_api/handlers.py
```diff
@@ -26,6 +26,10 @@ def default(self, o):
 def post_jobs(body, user):
     print(body)
 
+    if environ['SYSTEM_AVAILABLE'] != "true":
+        message = 'HyP3 is currently unavailable. Please try again later.'
```
Jlrine2, line 30: language?

api/tests/test_submit_job.py
```diff
@@ -99,3 +99,15 @@ def test_submit_job_granule_does_not_exist(client, table):
     assert response.json['title'] == 'Bad Request'
     assert response.json['detail'] == 'Some requested scenes could not be found: ' \
                                       'S1A_IW_SLC__1SDV_20200610T173646_20200610T173704_032958_03D14C_5F2A'
+
+
+def test_submit_job_hyp3_unavailable(client):
+    batch = [
+        make_job()
+    ]
```
asjohnston-asf, line 107: I think `submit_batch` creates this by default, can just call `submit_batch(client)`

PR #200, opened 2020-08-03
Add name field to jobs 

Jlrine2:
Name can be set in POST /jobs in the same way as description except for max length of 50 (arbitrary)
Name can be filtered in GET /jobs with query parameter name as exact match


api/src/hyp3_api/openapi-spec.yml
```diff
@@ -233,6 +242,13 @@ components:
       minLength: 1
       example: a description of the job
 
+    name:
+      description: User provided text to name the job
+      type: string
+      minLength: 1
+      maxLength: 50
```
Jlrine2, line 249: this is absolutely arbitrary, not sure if required or what it should actually be?

asjohnston-asf: I don't see a limit defined for subscription names, which is the closest parallel in v1

asjohnston-asf: 2609 total v1 subscriptions
all have a name, average length of 25 characters, median 17, max 119
1135 have a description, average length 93 characters, median 48, max 791

Jlrine2: I think I would rather get this number after the design dive, if the intent is to display this in vertex, the width of the container might indicate the appropriate limit.

PR #206, opened 2020-08-06
RELEASE: change browse image identification to use tags

Jlrine2:
* Files now get uploaded to S3 with a `file_type` tag indicating what the file is allowing us to use it in `get_files.py`


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [0.3.5]
```
jhkennedy, line 8: This will also release customization options to the API, but it's not listed in the CHANGELOG. Is that by design?

PR #215, opened 2020-08-07
Add failure path to get-files

jhkennedy:
Right now, if GET_FILES fails in the step function, the jobs are left in a RUNNING state forever. 

I *think* this is right...

### DIFF
--- a/step-function/cloudformation.yml
+++ b/step-function/cloudformation.yml
@@ -75,6 +75,11 @@ Resources:
                 "ErrorEquals": [ "States.ALL" ],
                 "MaxAttempts": 2
               }],
+              "Catch": [{
+                "ErrorEquals": ["States.ALL"],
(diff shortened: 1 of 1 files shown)


step-function/cloudformation.yml
```diff
@@ -74,6 +74,11 @@ Resources:
...
                 "ErrorEquals": [ "States.ALL" ],
                 "MaxAttempts": 2
+              }],
+              "Catch": [{
+                "ErrorEquals": ["States.ALL"],
+                "Next": "JOB_FAILED",
+                "ResultPath": "$.results.get_files"
               }],
                "ResultPath": "$.results.get_files",
                "Next": "JOB_SUCCEEDED"
```
asjohnston-asf, line 84: pre-existing issue, but should fix indentation for lines 83-85

PR #225, opened 2020-08-18
Release: remove description and revise DEM coverage checks

asjohnston-asf:
This release will break the currently deployed version of https://search-hyp3.asf.alaska.edu until the Disco team transitions from the `description` field to the `name` field.  They're OK with that, we can just ship.

I'll start going through the production hyp3 dynamodb table and removing descriptions from existing jobs; they'll need to be purged before this release or job queries that include descriptions will throw http 500 errors.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,10 +15,16 @@ jobs:
 
       - run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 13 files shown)


asjohnston-asf:
I've removed all descriptions from production.  Will need to check again when this is merged.

PR #229, opened 2020-08-26
Resolve http 500 error when submitting jobs with resolution: 30.0

asjohnston-asf:


api/src/hyp3_api/util.py
```diff
@@ -39,3 +40,13 @@ def get_request_time_expression(start, end):
...
+
+
+def convert_floats_to_decimals(element):
+    if type(element) is float:
+        return Decimal(element)
+    if type(element) is list:
+        return [convert_floats_to_decimals(item) for item in element]
+    if type(element) is dict:
+        return {key: convert_floats_to_decimals(value) for key, value in element.items()}
+    return element
```
asjohnston-asf, line 52: alternatively, `return json.loads(json.dumps(element), parse_float=Decimal)`

PR #232, opened 2020-08-27
isolate code that is specific to rtc_gamma for clear plugin condition…

Jlrine2:
…al statements

### DIFF
--- a/api/src/hyp3_api/openapi-spec.yml
+++ b/api/src/hyp3_api/openapi-spec.yml
@@ -207,6 +207,10 @@ components:
         - RTC_GAMMA
 
     job_parameters:
+      oneOf:
+        - $ref: "#/components/schemas/rtc_gamma_job_parameters"
+
+    rtc_gamma_job_parameters:
       description: Options and parameters for running the job process.
       type: object
       required:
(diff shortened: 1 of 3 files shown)


api/src/hyp3_api/util.py
```diff
@@ -8,6 +8,15 @@
 from hyp3_api import handlers
 
 
+def get_granules(body):
```
asjohnston-asf, line 11: this can take `jobs` as input, instead of getting the entire `body`

api/src/hyp3_api/util.py
```diff
@@ -8,6 +8,15 @@
...
+    granules = []
```
asjohnston-asf, line 12: could make this a set instead of a list and only return distinct granules

api/src/hyp3_api/openapi-spec.yml
```diff
@@ -207,6 +207,10 @@ components:
         - RTC_GAMMA
 
     job_parameters:
+      oneOf:
+        - $ref: "#/components/schemas/rtc_gamma_job_parameters"
```
asjohnston-asf, line 211: ideally we'd enforce that "`job_type": "RTC_GAMMA"` goes with `rtc_gamma_job_parameters and `"job_type": "INSAR_GAMMA" goes with `insar_gamma_job_parameters`, but I'm comfortable punting that until later

PR #239, opened 2020-09-02
take granules as list for all job types

Jlrine2:


api/src/hyp3_api/openapi-spec.yml
```diff
@@ -263,10 +261,18 @@ components:
         include_inc_map:
           $ref: "#/components/schemas/include_inc_map"
 
-    granule:
-      oneOf:
-        - $ref: "#/components/schemas/granule_slc"
-        - $ref: "#/components/schemas/granule_grdh"
+    insar_gamma_granules:
+      type: array
```
asjohnston-asf, line 264: length

api/src/hyp3_api/openapi-spec.yml
```diff
@@ -263,10 +261,18 @@ components:
...
+      items:
+        $ref: '#/components/schemas/granule_slc'
+
+    rtc_gamma_granules:
+      type: array
```
asjohnston-asf, line 269: length

asjohnston-asf:
`new_job` example needs to be updated  https://github.com/ASFHyP3/hyp3/blob/35ae911890651e8b2f9dbb4ed0fb9a0e711a00b5/api/src/hyp3_api/openapi-spec.yml#L141

asjohnston-asf:
and api.util.get_granules() needs updating: https://github.com/ASFHyP3/hyp3/blob/develop/api/src/hyp3_api/util.py#L11

api/src/hyp3_api/util.py
```diff
@@ -11,12 +11,7 @@
 def get_granules(jobs):
     granules = set()
     for job in jobs:
-        parameters = job['job_parameters']
-        if job['job_type'] == 'RTC_GAMMA':
-            granules.add(parameters['granule'])
-        if job['job_type'] == 'INSAR_GAMMA':
-            granules.add(parameters['granule1'])
-            granules.add(parameters['granule2'])
+        [granules.add(granule) for granule in job['job_parameters']['granules']]
```
asjohnston-asf, line 14: could be one big set comprehension, I think, but don't think I like it any better
`{granule for granule in job['job_parameters']['granules'] for job in jobs}`

and this assumes every job will have a `granules` parameter, but that's fine

api/tests/conftest.py
```diff
@@ -107,7 +107,8 @@ def make_db_record(job_id,
 
 
 def setup_requests_mock(batch):
-    granules = [job['job_parameters']['granule'] for job in batch]
+    granules = []
+    [[granules.append(granule) for granule in job['job_parameters']['granules']] for job in batch]
```
asjohnston-asf, line 111: this is a list comprehension, but we don't assign it to a variable?  odd way to implement a loop, I think it can be:
`granules = [granule for granule in job['job_parameters']['granules'] for job in batch]`

api/src/hyp3_api/openapi-spec.yml
```diff
@@ -215,22 +216,19 @@ components:
 #       - INSAR_GAMMA
```
asjohnston-asf, line 216: might as well re-enable INSAR_GAMMA at this point, although that might mean some additional test cases, so could do it as a separate PR

asjohnston-asf (approved):

PR #247, opened 2020-09-06
upgrade to v2 http api gateway

asjohnston-asf:
https://aws.amazon.com/blogs/compute/announcing-http-apis-for-amazon-api-gateway/
https://aws.amazon.com/blogs/compute/building-better-apis-http-apis-now-generally-available/
https://blog.thundra.io/api-gateway-gets-a-second-version

v2 HTTP solves many of our pain points with v1 REST.  Features we're not using are stripped away for lower latency and fewer cloudformation resources.  It also autodeploys to the `$default` stage with each stack update, which solves my biggest headache with v1.

I don't *think* our `cloudformation-deployment-role` needs any permission updates; looks like v1 and v2 both fall under `apigateway:*` actions.

The `serverless-wsgi` package we're using to translate between Api Gateway and WSGI only speaks [Payload Format Version](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html) 1.0, not 2.0, hence the override in the new `ApiGatewayManagedOverrides` resource.

v2 comes with simple CORs configuration as well.  We could consider moving CORs from python to cloudformation; it would be less code and we should only need it in a "real world" deployment.  I haven't pulled the trigger on that here, though.

### DIFF
--- a/api/cloudformation.yml
+++ b/api/cloudformation.yml
@@ -24,92 +24,48 @@ Parameters:
 Outputs:
 
   Url:
-    Value: !Sub "https://${DomainName}/ui"
+    Value: !Sub "https://${CustomDomainName}/ui"
 
 Resources:
 
-  RestApi:
-    Type: AWS::ApiGateway::RestApi
+  Api:
+    Type: AWS::ApiGatewayV2::Api
     Properties:
       Name: !Ref AWS::StackName
-      EndpointConfiguration:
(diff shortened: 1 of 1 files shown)


jhkennedy (approved):
Nice! This looks a lot leaner

Jlrine2 (approved):

PR #249, opened 2020-09-08
add list of names to user endpoint

Jlrine2:


jhkennedy (approved):

asjohnston-asf:
user schema needs a new property in the API spec: https://github.com/ASFHyP3/hyp3/blob/develop/api/src/hyp3_api/openapi-spec.yml#L98

api/tests/test_get_user.py
```diff
@@ -28,6 +28,10 @@ def test_get_user(client, table, monkeypatch):
             'limit': 25,
             'remaining': 21,
         },
+        'names': sorted([
```
asjohnston-asf, line 31: don't need `sorted` here, removing it verifies that the API is doing the sorting correctly

api/src/hyp3_api/handlers.py
```diff
@@ -90,13 +90,25 @@ def get_jobs(user, start=None, end=None, status_code=None, name=None):
...
+
+
 def get_user(user):
     return {
         'user_id': user,
         'quota': {
             'limit': int(environ['MONTHLY_JOB_QUOTA_PER_USER']),
             'remaining': get_remaining_jobs_for_user(user),
         },
+        'names': get_names_for_user(user)
```
asjohnston-asf, line 111: `user.job_names` is clearer that we're talking about jobs than `user.names`.  That might be a smell that we're coopting `/user` for something that might deserve a separate endpoint, but I'm fine with just clarifying the field name.

asjohnston-asf, line 111: `get_remaining_jobs_for_user` and `get_names_for_user` do very similar dynamodb queries, we could look at doing a single query and building both fields from the results, but it's only a performance issue and I'm not too worried about it right now

api/src/hyp3_api/handlers.py
```diff
@@ -90,13 +90,25 @@ def get_jobs(user, start=None, end=None, status_code=None, name=None):
     return {'jobs': response['Items']}
 
 
+def get_names_for_user(user):
+    table = DYNAMODB_RESOURCE.Table(environ['TABLE_NAME'])
+    key_expression = Key('user_id').eq(user)
+    response = table.query(
```
asjohnston-asf, line 96: need pagination here sooner or later

api/tests/test_get_user.py
```diff
@@ -11,10 +11,10 @@ def test_get_user(client, table, monkeypatch):
...
     user = 'user_with_jobs'
     items = [
-        make_db_record('job1', user_id=user, request_time=request_time, status_code='PENDING'),
-        make_db_record('job2', user_id=user, request_time=request_time, status_code='RUNNING'),
-        make_db_record('job3', user_id=user, request_time=request_time, status_code='FAILED'),
-        make_db_record('job4', user_id=user, request_time=request_time, status_code='SUCCEEDED')
+        make_db_record('job1', user_id=user, request_time=request_time, status_code='PENDING', name='job1'),
+        make_db_record('job2', user_id=user, request_time=request_time, status_code='RUNNING', name='job1'),
+        make_db_record('job3', user_id=user, request_time=request_time, status_code='FAILED', name='job2'),
+        make_db_record('job4', user_id=user, request_time=request_time, status_code='SUCCEEDED', name='job1')
```
asjohnston-asf, line 17: might do `name=None` for one of these?  don't know if it's a helpful test case

asjohnston-asf (approved):

PR #266, opened 2020-09-18
Use Jinja templating to create Cloud resource definitions

Jlrine2:
changes the structure of the repo so that CF templates and the JSON template can be created based on one file `job_types.yml` that defines different plugins


job_types.yml
```diff
@@ -0,0 +1,86 @@
...
+    - Ref::dem_matching
+    - --include-dem
+    - Ref::include_dem
+    - --include-inc-map
+    - Ref::include_inc_map
+    - Ref::granules
+  timeout: 5400
+
+InsarGamma:
+  image: 626226570674.dkr.ecr.us-east-1.amazonaws.com/hyp3-rtc-gamma
```
asjohnston-asf, line 40: wrong image

job_types.yml
```diff
@@ -0,0 +1,86 @@
...
+    - Ref::include_inc_map
+    - --include-los-displacement
+    - Ref::include_los_displacement
+    - --looks
+    - Ref::looks
+    - Ref::granules
+  timeout: 5400
+
+Autorift:
+  image: 626226570674.dkr.ecr.us-east-1.amazonaws.com/hyp3-rtc-gamma
```
asjohnston-asf, line 69: wrong image

.github/workflows/deploy.yml
```diff
@@ -33,6 +30,7 @@ jobs:
           fetch-depth: 0
           token: ${{ secrets.TOOLS_BOT_PAK }}
 
+
```
asjohnston-asf, line 33: drop extra whitespace

.github/workflows/deploy.yml
```diff
@@ -55,17 +53,18 @@ jobs:
         with:
           python-version: 3.8
 
-      - name: install dependancies
-        working-directory: api
+      - name: install dependancies and render templates
```
asjohnston-asf, line 56: `dependencies`

asjohnston-asf (approved):

PR #285, opened 2020-10-01
Add monitoring to hyp3

Jlrine2:


apps/monitoring-cf.yml
```diff
@@ -0,0 +1,51 @@
...
+    Properties:
+      AlarmActions:
+        - !Ref AlarmTopic
+      AlarmDescription: Hyp3 processing failures
+      ComparisonOperator: GreaterThanThreshold
+      Period: 60
+      EvaluationPeriods: 1
+      Threshold: 0
+      Dimensions:
+        - Name: Processing Failures
```
asjohnston-asf, line 26: `Name` is the AWS-defined dimension to filter by, I believe `StateMachineArn` in this case

apps/monitoring-cf.yml
```diff
@@ -0,0 +1,51 @@
...
+    Properties:
+      AlarmActions:
+        - !Ref AlarmTopic
+      AlarmDescription: Hyp3 api HTTP 500 errors
+      ComparisonOperator: GreaterThanThreshold
+      Period: 60
+      EvaluationPeriods: 1
+      Threshold: 0
+      Dimensions:
+        - Name: API 500 errors
```
asjohnston-asf, line 45: `ApiId` I think

apps/monitoring-cf.yml
```diff
@@ -0,0 +1,51 @@
...
+          Value: !Ref ApiId
+      MetricName: 5xxError
```
asjohnston-asf, line 47: just `5xx` I believe

asjohnston-asf:
I think you can pick out the namespace/metric/dimension values by building the metric manually in Cloudwatch, then looking at `Details`, I think it's `<namespace> * <metric name> * <dimension name>: <dimension value>`

Screenshot from 2020-09-30 17-28-53

asjohnston-asf:
We'll need to grant `cloudformation-deployment-role` additional permissions so it can manage alarms and topics, `sns:*` and `cloudwatch:*` would be sufficient

apps/monitoring-cf.yml
Jlrine2: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-metrics-and-dimensions.html
I think the console displays a simplified version

Jlrine2: although this documention reveals that our demention isn't valid

Jlrine2:
APi alarm works now

Jlrine2:
updated cloudformation-deployment-role

asjohnston-asf (approved):

PR #288, opened 2020-10-02
added user by user quotas definined in userTable

Jlrine2:


jhkennedy (approved):

apps/api/src/hyp3_api/handlers.py
```diff
@@ -103,11 +103,17 @@ def get_names_for_user(user):
 
 
 def get_user(user):
+    table = DYNAMODB_RESOURCE.Table(environ['USER_TABLE_NAME'])
+    response = table.get_item(Key={'user_id': user})
+    if 'Item' in response:
+        limit = response['Item']['max_jobs_per_month']
+    else:
+        limit = int(environ['MONTHLY_JOB_QUOTA_PER_USER'])
```
asjohnston-asf, line 111: this should all be delegated to a `get_quota_for_user` function

apps/api/api-cf.yml
```diff
@@ -3,6 +3,9 @@ Parameters:
   JobsTable:
     Type: String
 
+  UserTable:
```
asjohnston-asf, line 6: We have a mix of singular/plural.  Should be user/job everywhere or users/jobs everywhere, I don't have a strong opinion which.

tests/api/conftest.py
```diff
@@ -27,22 +27,25 @@ def client():
 
 
 @pytest.fixture
-def table():
-    table_properties = get_table_properties_from_template()
+def tables():
     with mock_dynamodb2():
```
asjohnston-asf, line 31: I slightly prefer separate `job_table` and `user_table` mocks, we might only have 1 test that uses both

asjohnston-asf (approved):

PR #289, opened 2020-10-02
GET /jobs/{job_id} API endpoint to search for a job by its job_id

jhkennedy:


Jlrine2 (approved):

apps/api/src/hyp3_api/handlers.py
```diff
@@ -91,6 +91,14 @@ def get_jobs(user, start=None, end=None, status_code=None, name=None):
     return {'jobs': response['Items']}
 
 
+def get_job_id(job_id):
```
asjohnston-asf, line 94: `get_job_by_id` is better

apps/api/src/hyp3_api/handlers.py
```diff
@@ -91,6 +91,14 @@ def get_jobs(user, start=None, end=None, status_code=None, name=None):
...
+    table = DYNAMODB_RESOURCE.Table(environ['TABLE_NAME'])
+    response = table.get_item(Key={'job_id': job_id})
+    if 'Item' not in response:
+        return problem(400, 'Bad Request', f'job_id does not exist: {job_id}')
```
asjohnston-asf, line 98: I think a `404` response might be better?

tests/api/test_get_job_id.py
```diff
@@ -0,0 +1,24 @@
...
+
+    job_id = '0ddaeb98-7636-494d-9496-03ea4a7df200'
+    response = client.get(f'{JOBS_URI}/{job_id}')
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
+    assert job_id in response.json['detail']
+
+    job_id = 'boogers'
+    response = client.get(f'{JOBS_URI}/{job_id}')
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
+    assert job_id in response.json['detail']
```
asjohnston-asf, line 24: could check that `jobs/` without an id returns 400, but that might be overkill

asjohnston-asf (approved):

PR #314, opened 2020-12-03
add s2 support to openapi

Jlrine2:


apps/api/src/hyp3_api/api-spec/job_parameters.yml.j2
```diff
@@ -55,6 +64,14 @@ components:
       maxLength: 67
       example: S1A_IW_SLC__1SSV_20150621T120220_20150621T120232_006471_008934_72D8
 
+    granule_s2:
+      description: The name of the Sentinel-2 SLC granule to process
+      type: string
+      pattern: "^S2[AB]_.*"
+      minLength: 67
```
jhkennedy, line 71: Setting `minLength: 24` here would allow both ESA and COG identifiers through
* COG example: `S2B_48NYN_20201203_0_L2A`

Both identifiers will be supported with ASFHyP3/hyp3-autorift#33

Otherwise, if we want strict length checking, we might need a `granule_s2_esa` and a `granule_s2_cog`

jhkennedy (changes requested):

jhkennedy (approved):

PR #317, opened 2020-12-04
fix s2 granule not found

Jlrine2:


apps/api/src/hyp3_api/validation.py
```diff
@@ -60,6 +60,7 @@ def get_cmr_metadata(granules):
 def check_granules_exist(granules, granule_metadata):
     found_granules = [granule['name'] for granule in granule_metadata]
     not_found_granules = set(granules) - set(found_granules)
+    not_found_granules = {granule for granule in not_found_granules if not granule.startwith('s2')}
```
asjohnston-asf, line 63: is this any clearer?
```
s2_granules = {granule for granule in not_found_granules if granule.startwith('s2')}
not_found_granules = set(granules) - set(found_granules) - s2_granules
```

asjohnston-asf (approved):

tests/api/test_validation.py
```diff
@@ -118,6 +118,7 @@ def test_check_granules_exist():
 
     with raises(GranuleValidationError) as e:
         check_granules_exist(['scene1', 'scene2', 'scene3', 'scene4'], granule_metadata)
```
asjohnston-asf, line 120: add `S2_this_should_not_be_found` to the list here

PR #327, opened 2020-12-16
Pull autorift container from west

jhkennedy:
Needs actions from to be complete ASFHyP3/hyp3-autorift#44

### DIFF
--- a/job_types.yml
+++ b/job_types.yml
@@ -108,7 +108,7 @@ InsarGamma:
   timeout: 5400
 
 Autorift:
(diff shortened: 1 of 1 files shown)


asjohnston-asf (approved):
Approved, once the new image is pushed after merging https://github.com/ASFHyP3/hyp3-autorift/pull/44

The autorift account will need to be granted pull permissions on 845172464411.dkr.ecr.us-west-2.amazonaws.com/hyp3-autorift before this change is released to production

asjohnston-asf:
Repo permissions have been added to allow the autorift account to pull images.

PR #347, opened 2021-02-18
up max jobs per POST /jobs request to 1000

asjohnston-asf:
I think this is the most expedient way to move forward in the short term.  We should work with the Vertex team on a better long term strategy.

### DIFF
--- a/apps/api/src/hyp3_api/api-spec/openapi-spec.yml
+++ b/apps/api/src/hyp3_api/api-spec/openapi-spec.yml
@@ -154,7 +154,7 @@ components:
       description: Contains a list of new job objects.
       type: array
       minItems: 1
-      maxItems: 200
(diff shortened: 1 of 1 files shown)


apps/api/src/hyp3_api/api-spec/openapi-spec.yml
```diff
@@ -154,7 +154,7 @@ components:
       description: Contains a list of new job objects.
       type: array
       minItems: 1
-      maxItems: 200
+      maxItems: 1000
       items:
```
Jlrine2, line 158: There are valid and unanswered questions about doing this, I think it would be okay to send it to test, but without knowing the extent to which this will effect latency I'm hesitant to send this change to production

Jlrine2 (approved):

PR #369, opened 2021-03-02
Allow list bucket for HyP3 content buckets

jhkennedy:
For uploading Unzipped files as well

### DIFF
--- a/apps/main-cf.yml
+++ b/apps/main-cf.yml
@@ -162,7 +162,7 @@ Resources:
           - Effect: Allow
             Principal: "*"
             Action: s3:ListBucket
-            Resource: !Sub "${ContentBucket.Arn}"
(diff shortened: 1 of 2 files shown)


apps/main-cf.yml
```diff
@@ -159,6 +159,10 @@ Resources:
             Principal: "*"
             Action: s3:GetObject
             Resource: !Sub "${ContentBucket.Arn}/*"
+          - Effect: Allow
+            Principal: "*"
+            Action: s3:ListBucket
+            Resource: !Sub "${ContentBucket.Arn}"
```
asjohnston-asf, line 165: `!GetAtt ContentBucket.Arn`

asjohnston-asf (approved):

PR #370, opened 2021-03-02
Have get-files ignore side-loaded product files (should be temporary!)

jhkennedy:
For side-loading zipped product files without telling anyone

### DIFF
--- a/apps/get-files/src/get_files.py
+++ b/apps/get-files/src/get_files.py
@@ -29,7 +29,7 @@ def get_object_file_type(bucket, key):
     return None
 
 
-def visible_product(product_path: Union[str, Path]):
+def visible_product(product_path: Union[str, Path]) -> bool:
     return Path(product_path).suffix in ('.zip', '.nc')
 
 
--- a/apps/get-files/src/get_files.py
+++ b/apps/get-files/src/get_files.py
@@ -1,6 +1,8 @@
 from datetime import datetime
 from os import environ
-from os.path import basename, splitext
+from os.path import basename
+from pathlib import Path
+from typing import Union
 
 import boto3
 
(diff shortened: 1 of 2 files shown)


apps/get-files/src/get_files.py
```diff
@@ -33,7 +33,7 @@ def get_products(files):
         'size': item['size'],
         'filename': item['filename'],
         's3': item['s3'],
-    } for item in files if item['file_type'] == 'product']
+    } for item in files if item['file_type'] == 'product' and splitext(item['filename'])[-1] in ('.zip', '.nc')]
```
asjohnston-asf, line 36: maybe `and Path(item['filename']).suffix in ('.zip', '.nc')`?

asjohnston-asf: or even:
```
def is_product(filename, file_type):
    return file_type == 'product' and Path(filename) in ('.zip', '.nc)
 
.... for item in files if is_product(item['filename'], item['file_type'])

**Review by asjohnston-asf — APPROVED** [review]

File: apps/get-files/src/get_files.py (hunk 1 above)
Comment (jhkennedy) [reply apps/get-files/src/get_files.py:36]: I originally did it with splitext because other parts of this code are `os.path` based -- `pathlib` isn't being used. 

I also left if a little dirty to intentionally be obnoxious as we shouldn't persist this change. 

I've changed it to a `visible_product` helper function with `pathlib`. I've settled on rather having a cleaner "dirty" implementation in case it actually persists much longer than it should.

File: apps/get-files/src/get_files.py [hunk 2]
```diff
@@ -27,13 +29,17 @@ def get_object_file_type(bucket, key):
     return None
 
 
+def visible_product(product_path: Union[str, Path]) -> bool:
```
asjohnston-asf, line 32: deserves a couple trivial tests?

jhkennedy: may as well

asjohnston-asf (approved):

PR #378, opened 2021-03-19
allow truncating results and start_tokens

Jlrine2:


asjohnston-asf:
Look the same as what I was envisioning, no concerns.

We'll also want to document the `next_token` field in the `GET /jobs` response spec at https://github.com/ASFHyP3/hyp3/blob/paging/apps/api/src/hyp3_api/api-spec/openapi-spec.yml#L110

apps/api/src/hyp3_api/handlers.py
```diff
@@ -82,9 +85,33 @@ def post_jobs(body, user):
     return body
 
 
-def get_jobs(user, start=None, end=None, status_code=None, name=None):
-    jobs = dynamo.query_jobs(user, start, end, status_code, name)
-    return {'jobs': jobs}
+def build_next_token(next_token):
```
asjohnston-asf: `build_next_token` seems like a weird name when you're passing in `next_token`, maybe `encode_token` or `encode_next_token`

tests/api/test_util.py
```diff
@@ -0,0 +1,21 @@
+from hyp3_api import util
+
+
+def test_encode_start_token():
+    token = {'foo': 1, 'bar': 2}
```
asjohnston-asf, line 5: maybe `payload` for the dictionaries

tests/api/test_util.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def test_token_handlers_invertable():
+    token = {'foo': 1, 'bar': 2}
+    assert util.deserialize(util.serialize(token)) == token
+
+
+def test_build_tokenized_url():
+    url = util.set_start_token('https://example.com/path?q1=foo&q2=bar', 'start_here')
+    assert url == 'https://example.com/path?q1=foo&q2=bar&start_token=start_here'
```
asjohnston-asf, line 21: should also test against a url that already has a `start_token=...`

asjohnston-asf (approved):

PR #405, opened 2021-04-29
add water mask option for insar gamma

Jlrine2:


Jlrine2:
@talogan This PR is a good example of adding a parameter to the HyP3 API, also do you have feedback on what should be in the description?

jhkennedy (approved):

asjohnston-asf (approved):

PR #432, opened 2021-05-29
correct water map parameter name in job_types.yml

asjohnston-asf:


jhkennedy (approved):
Weird, I know we corrected this at least once...

PR #437, opened 2021-06-03
Implement autoRIFT optimization strategy

jhkennedy:
This brings in the results of optimizing the memory usage for optical autoRIFT jobs (focused on Landsat-8 optimization).

This sets a step function up like: 
image


apps/workflow-cf.yml.j2
```diff
@@ -40,7 +40,7 @@ Resources:
         {% endfor %}
       ContainerProperties:
         Image: !Sub "{{ job_spec['image'] }}:${ImageTag}"
-        Vcpus: 4
+        Vcpus: 1
```
jhkennedy, line 43: Does/could this affect RTC and InSAR jobs?

asjohnston-asf: No.  Vcpus is a minimum requirement (launch a machine with at least this many CPUs), but not a hard cap like it is with memory.  A job with `Vcpus: 1` and `Memory: 30000` will run on an `r5d.2xlarge` and have access to all four CPUs.

asjohnston-asf (approved):

PR #445, opened 2021-06-07
modify ec2 instance type, add 10.0 as resolution option for RTC_GAMMA…

asjohnston-asf:
…, WATER_MAP jobs

Job timeouts in job_types.yml may need to be adjusted based on running a batch of sample 10m jobs.  Currently RTC_GAMMA has a timeout of 3 hours and WATER_MAP has a timeout of 6 hours.


jhkennedy (approved):
Lets see what the times look like, and bump as necessary

PR #449, opened 2021-06-08
increase watermap timeout to 24 hours

asjohnston-asf:
First ten 10m watermap jobs for random S1 granules all timed out after six hours.  The corresponding RTC jobs took between 45 minutes and 2 hours.


jhkennedy:
This is kinda ridiculous to need to do, but yeah!

PR #456, opened 2021-06-21
Remove connexion

Jlrine2:


apps/api/requirements-api.txt
```diff
@@ -1,7 +1,7 @@
-connexion[swagger-ui]
 cryptography
 flask
 Flask-Cors
+openapi-core
```
Jlrine2, line 4: https://openapi-core.readthedocs.io/en/latest/

also has Flask integration
https://openapi-core.readthedocs.io/en/latest/integrations.html#flask

apps/api/src/hyp3_api/handlers.py
```diff
@@ -34,39 +31,21 @@ def is_uuid(val):
...
-
 def post_jobs(body, user):
     print(body)
 
     monthly_quota = get_max_jobs_per_month(user)
     remaining_jobs = util.get_remaining_jobs_for_user(user, monthly_quota)
     if remaining_jobs - len(body['jobs']) < 0:
         message = f'Your monthly quota is {monthly_quota} jobs. You have {remaining_jobs} jobs remaining.'
-        return problem(400, 'Bad Request', message)
+        abort(error(400, 'Bad Request', message))
```
Jlrine2, line 41: These could just be one call to error, and that function could call abort, but I like that abort is here since it is more obvious that this will exit

jhkennedy (commented):
Overall looks fine to me...

asjohnston-asf:
I think we might be better served combining `templates/` and `static/` into a `ui/` directory. It eliminates the need to update some relative paths in `index.html`, and hopefully it's not much work to have out existing https://hyp3-api.asf.alaska.edu/ui/ route render `ui/index.html`.

Also, what version of swagger-ui was `static/` copied from? Diff is reporting differences between what's in this PR and the latest 3.50.0 release.

apps/api/src/hyp3_api/ui/README.md
```diff
@@ -0,0 +1,5 @@
+# [SwaggerUi](https://swagger.io/tools/swagger-ui/)
+
+The files in the swagger directory are from the Swagger UI [3.50.0](https://github.com/swagger-api/swagger-ui/releases/tag/v3.50.0) Release.
```
asjohnston-asf, line 3:
```suggestion
The files in the `ui/` directory are from the Swagger UI [3.50.0](https://github.com/swagger-api/swagger-ui/releases/tag/v3.50.0) Release.
```

apps/api/src/hyp3_api/ui/swagger/index.html
```diff
@@ -33,13 +33,13 @@
...
-    <script src="../static/swagger-ui-bundle.js" charset="UTF-8"> </script>
-    <script src="../static/swagger-ui-standalone-preset.js" charset="UTF-8"> </script>
+    <script src="./swagger-ui-bundle.js" charset="UTF-8"> </script>
+    <script src="./swagger-ui-standalone-preset.js" charset="UTF-8"> </script>
     <script>
     window.onload = function() {
       // Begin Swagger UI call region
       const ui = SwaggerUIBundle({
-        url: "{{ api_spec_url }}",
+        url: "openapi.json",
```
asjohnston-asf, line 42:
```suggestion
        url: "/openapi.json",
```

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [2.0.0](https://github.com/ASFHyP3/hyp3/compare/v1.1.6...v1.1.7)
+### Changed
+- Removed `connexion` library due to inactivity, replaced with `open-api-core`
+- Error messages from invalid input according to the `api-spec` are different
```
asjohnston-asf, line 11: The error messages note isn't specific enough to be useful. For example, what would a Vertex developer need to know to accommodate the change?

apps/api/src/hyp3_api/auth.py
```diff
@@ -4,7 +4,7 @@
 import jwt
 
 
-def decode_token(token, required_scopes):
+def decode_token(token):
     try:
         payload = jwt.decode(token, environ['AUTH_PUBLIC_KEY'], algorithms=environ['AUTH_ALGORITHM'])
         return {
```
asjohnston-asf, line 10: If we're no longer tied to the connexion implementation, I think we can simply `return payload` or `return payload['urs-user-id']` here

jhkennedy (approved):

PR #460, opened 2021-06-29
allow responses to "/ui/" to access the UI

Jlrine2:

### DIFF
--- a/apps/api/src/hyp3_api/routes.py
+++ b/apps/api/src/hyp3_api/routes.py
@@ -48,7 +48,7 @@ def authenticate_user():
 
 @app.route('/')
 def redirect_to_ui():
-    return redirect('/ui')
+    return redirect('/ui/')
 
 
 @app.route('/openapi.json')
--- a/apps/api/src/hyp3_api/routes.py
+++ b/apps/api/src/hyp3_api/routes.py
@@ -61,7 +61,7 @@ def get_open_api_yaml():
     return yaml.dump(api_spec_dict)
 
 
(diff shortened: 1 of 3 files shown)


tests/api/test_api_spec.py
```diff
@@ -98,3 +98,11 @@ def test_redirect_root(client):
     response = client.get('/')
     assert response.location.endswith('/ui')
     assert response.status_code == HTTPStatus.FOUND
+
+
+def test_ui_location(client):
+    response = client.get('/ui')
+    assert response.status_code == HTTPStatus.PERMANENT_REDIRECT
```
asjohnston-asf, line 105: should we assert the redirect location is `/ui/` as well?

asjohnston-asf:
```suggestion
    assert response.status_code == HTTPStatus.PERMANENT_REDIRECT
    assert response.location.endswith('/ui/')
```

apps/api/src/hyp3_api/routes.py
```diff
@@ -61,7 +61,7 @@ def get_open_api_yaml():
     return yaml.dump(api_spec_dict)
 
 
-@app.route('/ui')
+@app.route('/ui/')
```
asjohnston-asf, line 64: should update the `/` route to redirect to `/ui/` instead of `/ui` at https://github.com/ASFHyP3/hyp3/blob/develop/apps/api/src/hyp3_api/routes.py#L51

asjohnston-asf (approved):

PR #471, opened 2021-07-08
populate possible job_types in OpenAPI spec dynamically

asjohnston-asf:

### DIFF
--- a/apps/api/src/hyp3_api/api-spec/job_parameters.yml.j2
+++ b/apps/api/src/hyp3_api/api-spec/job_parameters.yml.j2
@@ -3,7 +3,7 @@ components:
     job_type:
       description: Type of process to run this job.
       type: string
-      example: RTC_GAMMA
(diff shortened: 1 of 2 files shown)


apps/api/src/hyp3_api/api-spec/job_parameters.yml.j2
```diff
@@ -1,5 +1,14 @@
 components:
   schemas:
+    job_type:
+      description: Type of process to run this job.
+      type: string
+      example: RTC_GAMMA
```
asjohnston-asf, line 6: Would like to populate this using something like `job_types.values()[0]['api_job_type']`, but haven't yet found a way to do it in the limited context of a jinja template

asjohnston-asf:
@AndrewPlayer3 FYI, this will introduce a merge conflict to be resolved if/when it's pulled into your `backprojection` branch.

Jlrine2 (approved):

PR #485, opened 2021-07-13
4 per machine for S2

jhkennedy:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [2.0.3](https://github.com/ASFHyP3/hyp3/compare/v2.0.2...v2.0.3)
+### Changed
+- Sentinel-2 autoRIFT jobs will utilize 3/4 less memory
```
asjohnston-asf, line 10: `utilize` suggests a reduction in the container itself, what about something like this?
```suggestion
- Sentinel-2 autoRIFT jobs will reserve 3/4 less memory, allowing more jobs to be run in parallel
```

asjohnston-asf (approved):

PR #489, opened 2021-07-15
Common lib

Jlrine2:


apps/api/requirements-api.txt
```diff
@@ -8,3 +8,4 @@ requests
 serverless_wsgi>=1.7.8
 shapely
 strict-rfc3339
+-e lib/dynamo
```
jhkennedy, line 11: I don't actually see this folder/library in the PR, did you push it?

Jlrine2: I did not

lib/dynamo/setup.py
```diff
@@ -0,0 +1,12 @@
...
+setup(
+    name='dynamo',
+
+    license='BSD',
+    include_package_data=True,
+
+    python_requires='~=3.8',
+
+    packages=find_packages(),
+)
```
jhkennedy, line 12: Probably should have a `install_requires=`

jhkennedy (approved):
This makes sense to me. `lib/dynamo/dynamo` implies more libraries are coming, which I think is probably a good decision.

PR #490, opened 2021-07-26
update description of include_inc_map api parameter

asjohnston-asf:


asjohnston-asf:
Hrm, we're using a common definition of the `include_inc_map` parameter for both InSAR and RTC jobs. Is there a more generic description that can cover both the RTC case (local only) and the InSAR case (local and ellipsoid)?

Jlrine2 (approved):

PR #501, opened 2021-08-03
Allow environment variable based customization of rendered yaml 

jhkennedy:
* adds [pyaml-env](https://pypi.org/project/pyaml-env/) as a dependency to allow `!ENV` directives
* uses `!ENV` to set the autoRIFT naming scheme
* expands the deployment action to set a space separated list of `key=value` pairs as environment variables

### DIFF
--- a/.github/actions/deploy-hyp3/action.yml
+++ b/.github/actions/deploy-hyp3/action.yml
@@ -36,15 +36,28 @@ inputs:
     description: "The default number of jobs any user with an Earthdata Login can run per month"
     required: false
     default: "0"
+  CUSTOM_ENV_VARS:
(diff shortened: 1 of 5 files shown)


Jlrine2:
annother potential option would be to split the rendered file into deployment specific files:
```
/config
   job-types.yml
   job-types-autorift.yml
   job-types-foo.yml
```

and use deploy time variables to decide which config to use, also would give us the oportunity to pull jobs out of autorift that don't need to be there and a mechanism to add jobs to one deployment without adding them to HyP3 prod and test

jhkennedy:
Closing for #502 -- this was an interesting attempt but 502 is more in line with how we're doing everything else, and we're scratching the surface of a larger structural problem so we should deep dive on this later.

PR #502, opened 2021-08-03
Add CloudFormation parameter for autoRIFT naming scheme

asjohnston-asf:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -18,6 +18,7 @@ jobs:
             image_tag: latest
             quota: 1000
             deploy_ref: refs/heads/main
+            autorift_naming_scheme: ITS_LIVE_OD
 
           - environment: hyp3-test
             domain: hyp3-test-api.asf.alaska.edu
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy.yml
```diff
@@ -64,6 +67,7 @@ jobs:
           EDL_PASSWORD: ${{ secrets.EDL_PASSWORD }}
           CLOUDFORMATION_ROLE_ARN: ${{ secrets.CLOUDFORMATION_ROLE_ARN }}
           MONTHLY_JOB_QUOTA_PER_USER: ${{ matrix.quota }}
+          AUTORIFT_NAMING_SCHEME: ${{ matrix.autorift_naming_scheme }}
```
jhkennedy, line 70: You'll have to add an input parameter to the deploy action to accept this variable:
https://github.com/ASFHyP3/hyp3/blob/naming-scheme-as-parameter/.github/actions/deploy-hyp3/action.yml#L4

jhkennedy (changes requested):

Jlrine2 (approved):

.github/workflows/deploy.yml
jhkennedy: Nevermind I'm a dummy and missed it.

jhkennedy (approved):

PR #512, opened 2021-08-05
populate derived job fields in dynamo.jobs.put_job instead of api.handlers

asjohnston-asf:


apps/api/src/hyp3_api/handlers.py
```diff
@@ -48,16 +47,8 @@ def post_jobs(body, user):
...
-    jobs = []
-    for job in body['jobs']:
-        job['job_id'] = str(uuid4())
-        job['user_id'] = user
-        job['status_code'] = 'PENDING'
-        job['request_time'] = request_time
-        jobs.append(util.convert_floats_to_decimals(job))
     if not body.get('validate_only'):
-        dynamo.jobs.put_jobs(jobs)
+        body['jobs'] = dynamo.jobs.put_jobs(user, body['jobs'])
```
asjohnston-asf, line 51: With this change the response when `validate_only=true` will not include the derived fields. I *think* that's an improvement, but I need to mull it a little still.

Jlrine2: I agree that this is an improvement

Jlrine2 (approved):

PR #529, opened 2021-08-17
Distro cors

Jlrine2:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [2.2.0](https://github.com/ASFHyP3/hyp3/compare/v2.2.1...v2.2.0)
+## [2.2.1](https://github.com/ASFHyP3/hyp3/compare/v2.2.0...v2.2.1)
+### Changed
+- HyP3 content bucket now allows Cross Origin Resource Headers
+
+## [2.2.0](https://github.com/ASFHyP3/hyp3/compare/v2.1.1...v2.2.0)
 ### Added
 - `GET` and `POST` `/subscriptions` endpoint which allows user to define a subscription with search and processing criteria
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,7 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [2.2.0](https://github.com/ASFHyP3/hyp3/compare/v2.2.1...v2.2.0)
+## [2.2.1](https://github.com/ASFHyP3/hyp3/compare/v2.2.0...v2.2.1)
```
asjohnston-asf, line 7: I expect this will be either part of the 2.2.0 release along with subscriptions, or possibly a 2.1.2 release prior to subscriptions

apps/main-cf.yml
```diff
@@ -167,6 +167,17 @@ Resources:
...
+        CorsRules:
+          - AllowedHeaders:
+              - '*'
+            AllowedMethods:
+              - GET
+              - HEAD
+            AllowedOrigins:
+              - null
+            ExposeHeaders: []
+            MaxAge: 3000
```
asjohnston-asf, line 180: Per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-cors-corsrule.html , only AllowedMethods and AllowedOrigins are required, so I'd err on the side of not specifying the rest unless we have a specific reason:
```suggestion
            AllowedMethods:
              - GET
              - HEAD
            AllowedOrigins:
              - "*"
```

Jlrine2: per brians advice the other 2 are required

CHANGELOG.md
Jlrine2: Fair, lets do 2.1.2, I think the workflow would be

Change the changelog, removing the subscriptions entry (2.2.0), merge to develop, restore branch, merge to main, re add subscriptions entry

Jlrine2: actually I've changed my mind, lets send it with subscriptions

apps/main-cf.yml
Jlrine2: `AllowedOrigins` also specifically has to be `null`

asjohnston-asf (approved):

PR #530, opened 2021-08-17
Update subscriptions

Jlrine2:


apps/api/src/hyp3_api/api-spec/openapi-spec.yml
```diff
@@ -161,6 +188,13 @@ components:
         job_specification:
           $ref: "#components/schemas/subscription_job_specification"
 
+    subscriptions_update_expression:
+      description: Change/update a subscription
+      type: object
+      properties:
```
asjohnston-asf, line 194: should include `additionalProperties: false`, and I lean towards making `end` required

tests/test_api/test_subscriptions.py
```diff
@@ -173,3 +173,92 @@ def test_get_subscriptions(client, tables):
     response = client.get(SUBSCRIPTIONS_URI)
     assert response.json == {'subscriptions': items}
     assert response.status_code == HTTPStatus.OK
+
```
asjohnston-asf, line 176: can add `test_get_subscription_by_id()`

asjohnston-asf (approved):

PR #552, opened 2021-09-01
add option to enable/disable subscriptions

Jlrine2:

### DIFF
--- a/apps/process-new-granules/src/process_new_granules.py
+++ b/apps/process-new-granules/src/process_new_granules.py
@@ -75,5 +75,5 @@ def lambda_handler(event, context):
         if subscription['enabled']:
             handle_subscription(subscription)
 
-            if end_filter <= dateutil.parser.parse(subscription['search_parameters']['end']):
+            if end_filter > dateutil.parser.parse(subscription['search_parameters']['end']):
                 disable_subscription(subscription)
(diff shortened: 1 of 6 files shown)


apps/process-new-granules/src/process_new_granules.py
```diff
@@ -67,5 +72,8 @@ def lambda_handler(event, context):
     subscriptions = dynamo.subscriptions.get_all_subscriptions()
     for subscription in subscriptions:
         end_filter = datetime.now(tz=timezone.utc) - timedelta(days=5)
-        if end_filter <= dateutil.parser.parse(subscription['search_parameters']['end']):
+        if subscription['enabled']:
             handle_subscription(subscription)
+
+            if end_filter <= dateutil.parser.parse(subscription['search_parameters']['end']):
+                disable_subscription(subscription)
```
jhkennedy, line 79: Do you want to do this before you handle subscription?

tests/test_process_new_granules.py
```diff
@@ -1,3 +1,4 @@
+from datetime import datetime, timezone, timedelta
```
asjohnston-asf, line 1:
```suggestion
from datetime import datetime, timedelta, timezone
```

asjohnston-asf (approved):

PR #567, opened 2021-09-14
add validate_only option to POST /subscriptions

Jlrine2:

### DIFF
--- a/apps/api/src/hyp3_api/routes.py
+++ b/apps/api/src/hyp3_api/routes.py
@@ -1,6 +1,5 @@
 import datetime
 import json
-from copy import deepcopy
 from decimal import Decimal
 from os import environ
 from pathlib import Path
--- a/apps/api/src/hyp3_api/routes.py
+++ b/apps/api/src/hyp3_api/routes.py
@@ -162,7 +162,10 @@ class Subscriptions(FlaskOpenAPIView):
 
     def post(self):
         body = request.get_json()
-        return jsonify(handlers.post_subscriptions(body, g.user))
+        validate_only = body.get('validate_only')
+        if validate_only is not None:
+            del body['validate_only']
+        return jsonify(handlers.post_subscriptions(body, g.user, validate_only))
 
     def get(self, subscription_id):
         if subscription_id is not None:
--- a/apps/api/src/hyp3_api/routes.py
+++ b/apps/api/src/hyp3_api/routes.py
@@ -163,9 +163,7 @@ class Subscriptions(FlaskOpenAPIView):
 
     def post(self):
         body = request.get_json()
-        validate_only = deepcopy(body.get('validate_only'))
-        if validate_only is not None:
-            del body['validate_only']
+        validate_only = body.pop('validate_only', False)
         return jsonify(handlers.post_subscriptions(body, g.user, validate_only))
 
     def get(self, subscription_id):
--- a/apps/api/src/hyp3_api/routes.py
+++ b/apps/api/src/hyp3_api/routes.py
@@ -162,8 +162,7 @@ class Subscriptions(FlaskOpenAPIView):
 
     def post(self):
         body = request.get_json()
-        validate_only = body.pop('validate_only', False)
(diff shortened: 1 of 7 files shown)


tests/test_dynamo/test_subscriptions.py
```diff
@@ -214,3 +214,38 @@ def test_put_subscription_update(tables):
 
     response = tables.subscriptions_table.scan()
     assert response['Items'] == [updated_subscription]
+
+
+def test_put_subscription_validate_only(tables):
+    subscription = {
```
asjohnston-asf, line 220: `bad_subscription` might be more expressive here, and `good_subscription` later

tests/test_dynamo/test_subscriptions.py
```diff
@@ -214,3 +214,38 @@ def test_put_subscription_update(tables):
...
+            'job_type': 'RTC_GAMMA',
+            'name': 'sub1',
+        },
+        'search_parameters': {
+            'start': '2020-01-01T00:00:00+00:00',
+            'end': '2020-01-01T00:00:00+00:00',
+        }
+    }
+    with pytest.raises(ValueError):
+        dynamo.subscriptions.put_subscription('user1', subscription, validate_only=True)
```
asjohnston-asf, line 231: Checking for ValueError when validate_only=False is covered elsewhere, yes?

tests/test_dynamo/test_subscriptions.py
```diff
@@ -214,3 +214,38 @@ def test_put_subscription_update(tables):
...
+    }
+    dynamo.subscriptions.put_subscription('user1', subscription, validate_only=True)
+    assert tables.subscriptions_table.scan()['Items'] == []
+
+    dynamo.subscriptions.put_subscription('user1', subscription, validate_only=False)
+    assert tables.subscriptions_table.scan()['Items'] == [subscription]
+
+    tables.subscriptions_table.delete_item(Key={'subscription_id': subscription['subscription_id']})
+    dynamo.subscriptions.put_subscription('user1', subscription, validate_only=None)
+    assert tables.subscriptions_table.scan()['Items'] == [subscription]
```
asjohnston-asf, line 251: I think I'd strike this test case in favor of `validate_only` always taking a boolean, never None

apps/api/src/hyp3_api/routes.py
```diff
@@ -162,7 +163,10 @@ def __init__(self, spec):
 
     def post(self):
         body = request.get_json()
-        return jsonify(handlers.post_subscriptions(body, g.user))
+        validate_only = deepcopy(body.get('validate_only'))
+        if validate_only is not None:
+            del body['validate_only']
```
asjohnston-asf, line 168: this kind of logic is enough for me to want to move the `validate_only` key outside of the subscription object, e.g:
```
{
  "subscription": {
    "search_parameters": {...},
    "job_specification": {...}
  }
  "validate_only': false
}
```
but that's cumbersome for the 99.9% of cases where you're not going to bother specifying `validate_only`

tests/test_dynamo/test_subscriptions.py
asjohnston-asf: looks like ValueError is only checked for in `test_validate_subscription()`, and `validate_subscription()` doesn't know about validate_only, so I'd err on the side of adding the two exra lines to do the check here

apps/api/src/hyp3_api/routes.py
asjohnston-asf: otherwise `validate_only = body.pop('validate_only', False)` might be equivalent and more concise

tests/test_dynamo/test_subscriptions.py
Jlrine2: so nothing actionable?

asjohnston-asf: I'd add
```
    with pytest.raises(ValueError):
        dynamo.subscriptions.put_subscription('user1', subscription, validate_only=False)
```

jhkennedy (approved):

PR #568, opened 2021-09-15
Query subscriptions

Jlrine2:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `POST /subscriptions` requests may now include a `validate_only` key which when set to `true` will not add the subscription
 to the database but still validate it.
 - in `POST /subscriptions` requests, `search_parameters` and `job_specification` are now included under `subscription`
-- `GET /subscriptions` requests not include query parameters
+- `GET /subscriptions` requests now include query parameters
   - `name` gets only subscriptions with the given name
   - `job_type` gets only subscriptions with the given job type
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `POST /subscriptions` requests may now include a `validate_only` key which when set to `true` will not add the subscription
 to the database but still validate it.
 - in `POST /subscriptions` requests, `search_parameters` and `job_specification` are now included under `subscription`
+- `GET /subscriptions` requests now include query parameters
+  - `name` gets only subscriptions with the given name
```
jhkennedy, line 13: ...with the given project name? Job name?

apps/main-cf.yml
```diff
@@ -254,5 +256,7 @@ Resources:
           KeySchema:
             - AttributeName: user_id
               KeyType: HASH
+            - AttributeName: creation_date
+              KeyType: RANGE
           Projection:
             ProjectionType: ALL
```
jhkennedy, line 262: trailing new line?

jhkennedy (approved):
Looks good to me! Just two super minor things

apps/main-cf.yml
Jlrine2: didn't have one before

CHANGELOG.md
Jlrine2: I'm leaving it as `name` because that is the name of field in both subscriptions and jobs

jhkennedy: I think you should probably specify that it's in the job specification, not a subscription name

PR #578, opened 2021-09-20
prototype changing upload_log to be an activity

Jlrine2:


apps/upload-log/upload-log-cf.yml
```diff
@@ -7,11 +7,16 @@ Parameters:
...
-    Value: !GetAtt Lambda.Arn
+  UploadLogActivity:
+    Value: !Ref UploadLogActivity
 
 Resources:
 
+  UploadLogActivity:
+    Type: AWS::StepFunctions::Activity
+    Properties:
+      Name: !Sub "${AWS::StackName}-upload-log-activity"
```
asjohnston-asf, line 18: The lambda IAM role will need additional permissions to interact with the activity

apps/upload-log/upload-log-cf.yml
```diff
@@ -7,11 +7,16 @@ Parameters:
 
 Outputs:
 
-  LambdaArn:
-    Value: !GetAtt Lambda.Arn
+  UploadLogActivity:
```
asjohnston-asf, line 10: I'd consider calling this `ActivityArn`. `UploadLog` is probably redundant from context, and it specifies the value is a full ARN (as opposed to a name).

apps/upload-log/upload-log-cf.yml
asjohnston-asf: I expect `upload-log` will be in the stack name, so explicitly inluding it might be redundant
```suggestion
  UploadLogActivity:
    Type: AWS::StepFunctions::Activity
    Properties:
      Name: !Sub "${AWS::StackName}-activity"
```

PR #585, opened 2021-09-22
Makefile

Jlrine2:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,14 +13,4 @@ jobs:
         with:
           python-version: 3.8
 
-      - run: |
-          python -m pip install --upgrade pip
-          python -m pip install -r requirements-all.txt
-          python apps/render_cf.py --job-types-file job_types.yml
-
-      - name: run pytest
-        env:
-          PYTHONPATH: apps/api/src:apps/get-files/src:apps/start-execution/src:apps/upload-log/src:apps/process-new-granules/src
-        run: |
-          export $(xargs < tests/cfg.env)
-          pytest tests
+      - run: make test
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -13,14 +13,4 @@ jobs:
         with:
           python-version: 3.8
 
-      - run: |
-          python -m pip install --upgrade pip
```
jhkennedy, line 17: I think you'll want to leave this

Makefile
```diff
@@ -0,0 +1,21 @@
+API = ${PWD}/apps/api/src
+GET_FILES = ${PWD}/apps/get-files/src
+START_EXECUTION = ${PWD}/apps/start-execution/src
+UPLOAD_LOG = ${PWD}/apps/upload-log/src
+PROCESS_NEW_GRANULES = ${PWD}/apps/process-new-granules/src
+DYNAMO = ${PWD}/lib/dynamo
+export PYTHONPATH := ${API}:${GET_FILES}:${START_EXECUTION}:${UPLOAD_LOG}:${PROCESS_NEW_GRANULES}:${DYNAMO}
```
jhkennedy, line 7: I'm not sure we want the `:=` here -- if users already have a `PYTHONPATH` this won't do anything.

Might be better to do an `=` and blow away any user `PYTHONPATH` for these tests. 

This also requires you to be at the repository root, which I don't think is a problem

Makefile
```diff
@@ -0,0 +1,21 @@
...
+
+run: install render
+	export $$(xargs < tests/cfg.env); \
+	python apps/api/src/hyp3_api/__main__.py
+
+install:
+	python -m pip install -r requirements-all.txt
+
+render:
+	python apps/render_cf.py --job-types-file job_types.yml
```
jhkennedy, line 21: trailing newline

PR #611, opened 2021-10-12
update templating to define job parameter specs in job_types

Jlrine2:
Moves job types file definitions into job_spec folder
Move api job parameter definitions into job_spec folder


apps/api/src/hyp3_api/api-spec/job_parameters.yml.j2
```diff
@@ -27,178 +27,26 @@ components:
...
-    los_displacement:
-      description: Include a GeoTIFF in the product package containing displacement values along the Line-Of-Sight (LOS)
-      default: false
-      type: boolean
-
-    include_displacement_maps:
-      description: Include displacement maps (line-of-sight and vertical) in the product package
-      default: false
-      type: boolean
+    {% endfor %}
```
jhkennedy, line 52:
```suggestion
    {% endfor %}

```

apps/render_cf.py
```diff
@@ -33,12 +37,11 @@ def get_env():
...
 
 def main():
-    parser = ArgumentParser()
-    parser.add_argument('--job-types-file', required=True)
-    job_types_file = parser.parse_args().job_types_file
-
-    with open(job_types_file) as f:
-        job_types = yaml.safe_load(f)
+    job_types = {}
+    job_spec_path = Path('./job_spec')
```
jhkennedy, line 41: Could reference this from `__file__` instead of `cwd` ... might make it so you could render from anywhere

job_spec/INSAR_GAMMA.yml
```diff
@@ -0,0 +1,101 @@
...
+  - --include-wrapped-phase
+  - Ref::include_wrapped_phase
+  - --apply-water-mask
+  - Ref::apply_water_mask
+  - --looks
+  - Ref::looks
+  - Ref::granules
+validators:
+  - check_dem_coverage
+timeout: 10800
```
jhkennedy, line 101:
```suggestion
timeout: 10800

```

apps/render_cf.py
Jlrine2:
```suggestion
    job_spec_path = Path(__file__) / 'job_spec'
```

jhkennedy: I think you'll need `Path(__file__).parent / 'job_spec'`

PR #654, opened 2021-11-01
convert Deximals to floats before passing to search

Jlrine2:

### DIFF
--- a/apps/process-new-granules/src/process_new_granules.py
+++ b/apps/process-new-granules/src/process_new_granules.py
@@ -15,7 +15,7 @@ def get_unprocessed_granules(subscription):
     )
     processed_granules = [job['job_parameters']['granules'][0] for job in processed_jobs]
 
-    search_results = asf_search.search(**subscription['search_parameters'])
+    search_results = asf_search.search(**dynamo.util.convert_floats_to_decimals(subscription['search_parameters']))
     return [result for result in search_results if result.properties['sceneName'] not in processed_granules]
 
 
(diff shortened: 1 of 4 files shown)


apps/process-new-granules/src/process_new_granules.py
```diff
@@ -15,7 +15,7 @@ def get_unprocessed_granules(subscription):
     )
     processed_granules = [job['job_parameters']['granules'][0] for job in processed_jobs]
 
-    search_results = asf_search.search(**subscription['search_parameters'])
+    search_results = asf_search.search(**dynamo.util.convert_floats_to_decimals(subscription['search_parameters']))
```
jhkennedy, line 18: Should this be:
```suggestion
    search_results = asf_search.search(**dynamo.util.convert_decimals_to_floats(subscription['search_parameters']))
```

lib/dynamo/dynamo/util.py
```diff
@@ -36,3 +36,13 @@ def convert_floats_to_decimals(element):
     if type(element) is dict:
         return {key: convert_floats_to_decimals(value) for key, value in element.items()}
     return element
+
+
+def convert_decimals_to_floats(element):
+    if type(element) is Decimal:
+        return float(element)
```
asjohnston-asf, line 43: I suspect there are some values we would want to cast as `int` instead of float? I expect this issue was uncovered because a subscription included path/frame search parameters, which I would expect to pass to asf_search as `int`s?

asjohnston-asf: `asf_search.search` function definition: https://github.com/asfadmin/Discovery-asf_search/blob/master/asf_search/search/search.py#L39

asjohnston-asf (approved):

PR #661, opened 2021-11-04
Allow job spec YAMLs to specify an image tag

jhkennedy:
This PR allows job specification YAML's to override the default deployment image tag (`ImageTag` cloudformation deployment parameter) by including `image_tag` in the YAML like:
```yaml
INSAR_GAMMA:
  image: 845172464411.dkr.ecr.us-west-2.amazonaws.com/hyp3-gamma
  image_tag: '4.9.0'
  required_parameters:
    - granules
   ...
 ```
 
 We'veneeded this for the watermapping deployment to freeze the gamma plugin version, and would allow us to put both a prod and test job in a custom deployment for prototyping work

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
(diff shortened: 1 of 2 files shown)


jhkennedy:
I've opened #662 and #665, which shows how this option would/could be used.

 Unfortunately we can't get bring these deployments into the normal workflow because they break Vertex, so that's why I've done this as a PR into dev and two PRs that build off of it.

asjohnston-asf (approved):

PR #699, opened 2021-11-23
refactor scale_cluster.py

asjohnston-asf:

### DIFF
--- a/.github/actions/deploy-hyp3/action.yml
+++ b/.github/actions/deploy-hyp3/action.yml
@@ -45,7 +45,7 @@ inputs:
     description: "Comma separated list of IP CIDR blocks that will receive HTTP 403 FORBIDDEN responses from the API"
     required: true
   MONTHLY_COMPUTE_BUDGET:
-    description: "Total budget allocated for AWS Batch processing per month, in dollars"
+    description: "Total budget allocated for AWS Batch processing per month, in dollars. Ignored when DefaultMaxvCpus = ExpandedMaxvCpus."
     required: true
   DEFAULT_MAX_VCPUS:
     description: "Default maximum size for the AWS Batch compute environment"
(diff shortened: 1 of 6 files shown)


.github/actions/deploy-hyp3/action.yml
```diff
@@ -45,7 +45,7 @@ inputs:
     description: "Comma separated list of IP CIDR blocks that will receive HTTP 403 FORBIDDEN responses from the API"
     required: true
   MONTHLY_COMPUTE_BUDGET:
-    description: "Total budget allocated for AWS Batch processing per month, in dollars"
+    description: "Total budget allocated for AWS Batch processing per month, in dollars. Ignored when DefaultMaxvCpus = ExpandedMaxvCpus."
```
jhkennedy, line 48: Worth moving below `DEFAULT_MAX_VCPUS` and `EXPANDED_MAX_VCPUS`?

apps/scale-cluster/src/scale_cluster.py
```diff
@@ -33,33 +33,37 @@ def get_ec2_spending_month_to_date():
...
-    max_spending_for_next_day = (target_spending - spending)
+    print(f'Month-to-date compute budget: ${month_to_date_compute_budget:,.2f}')
+    print(f'Month-to-date compute spending: ${month_to_date_compute_spending:,.2f}')
+    print(f'Available surplus: ${available_surplus:,.2f}')
+    print(f'Required surplus: ${required_surplus:,.2f}')
 
-    if max_spending_for_next_day >= required_surplus:
-        max_vcpus = expanded_max_vcpus
-    else:
+    if available_surplus < required_surplus:
```
jhkennedy, line 55: With a binary switch between fleet sizes, I would almost rather return this (fn would be `use_expand_fleet` or something then), and keep the if-else logic in `lambda_handler` as it would drop two parameters from this function that are only used in the if-else. 

If we go to a more dynamic fleet size, I would want it so you get something between a zero and one back, and use that value to set the vcpus. Something like:
```python
fleet_size = default_max_vcpus + epansion_factor * (absolute_max_vcpus - default_max_vcpus)
```

jhkennedy (approved):
A few minor comments is all. 

A general one: I'm not sure I like `expanded_max_vcpus` and would probably prefer `absolute_max_vcpus` so that if would work if we switch to a more dynamic change to fleet size (instead of a binary switch in sizes).  But I'm not sure if we ever will.

apps/scale-cluster/src/scale_cluster.py
asjohnston-asf: "Default", "Expanded" and "Surplus" are the words I'd like to make more expressive, I'm open to suggestions to replace any/all of them.

asjohnston-asf:
"Default", "Expanded" and "Surplus" are the words I'd like to make more expressive, I'm open to suggestions to replace any/all of them.

PR #711, opened 2021-12-03
Make custom domain name and ssl certificate optional stack parameters

asjohnston-asf:
This prepares for deployment into AWS environments (JPL, Earthdata Cloud) where deploying custom domain names and SSL certificates will require coordination with an external platform security team.

API authentication is still dependent on cookies, which will continue to present a challenge when accessing the API at a domain name outside the `*.asf.alaska.edu` umbrella.

I'm considering whether we should [disable the default endpoint](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-api.html#cfn-apigatewayv2-api-disableexecuteapiendpoint) (e.g. https://abc123.execute-api.us-west-2.amazonaws.com)  when a custom domain name is provided. Any thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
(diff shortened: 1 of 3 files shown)


jhkennedy (approved):
Nice! Much better than commenting out for EDC/JPL deployments.

PR #725, opened 2021-12-07
Use reusable workflows

jhkennedy:
These actions are described in ASFHyP3/actions#3, which inlcudes working usage examples


.github/workflows/deploy.yml
```diff
@@ -169,16 +169,20 @@ jobs:
           owner: ASFHyP3
           name: hyp3
 
+      - name: Get tag data
+        run: |
+          LABEL_QUERY='.data.repository.pullRequest.labels.nodes[].name'
+          SELECT='select(. == "major" or . == "minor" or . == "patch")'
+          echo "BUMP_PART=$(jq --raw-output  "${LABEL_QUERY} | ${SELECT}" labels.json | sort | head -1)" >> $GITHUB_ENV
+          PR_QUERY='.data.repository.commit.associatedPullRequests.edges[0].node.title'
+          echo "TAG_MSG=$(jq --raw-output "${PR_QUERY}"  pr.json)" >> $GITHUB_ENV
```
asjohnston-asf, line 178: thoughts on making these part of the `tag version` step, since we don't technically need to fetch this information if `BUMP_PART == 'bumpless'`?

asjohnston-asf: or it's own `get tag message` step? might be a little easier to read that way, these five lines are pretty dense and I have to look down at the `tag version` step to figure out that `BUMP_PART` and `TAG_MSG` are the two outputs (and the only outputs)

jhkennedy: switching to a reusable workflow

asjohnston-asf (approved):

PR #731, opened 2021-12-08
use iam permissions boundary and vpc in edc

Jlrine2:

### DIFF
--- a/.github/actions/deploy-hyp3/action.yml
+++ b/.github/actions/deploy-hyp3/action.yml
@@ -61,7 +61,7 @@ inputs:
     required: true
   DEPLOY_LAMBDAS_IN_VPC:
     description: "Whether or not to deploy lambda functions in a VPC"
-    required: false
+    required: true
 
 runs:
   using: "composite"
--- a/.github/actions/deploy-hyp3/action.yml
+++ b/.github/actions/deploy-hyp3/action.yml
@@ -56,6 +56,12 @@ inputs:
   REQUIRED_SURPLUS:
     description: "Amount by which month-to-date budget must exceed month-to-date spending to increase fleet size, in dollars. Ignored when DefaultMaxvCpus = ExpandedMaxvCpus."
     required: true
+  PERMISSIONS_BOUNDARY_ARN:
+    description: "Arn of permissions boundary to provide to any created IAM roles"
+    required: true
+  DEPLOY_LAMDAS_IN_VPC:
+    description: "Whether or not to deploy lambda functions in a VPC"
+    required: false
 
 runs:
   using: "composite"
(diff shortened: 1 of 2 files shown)


asjohnston-asf:
need to hand the new parameters through to the github action at https://github.com/ASFHyP3/hyp3/blob/ec12fd3dbec35c5778538ba2601c0f5632ee503e/.github/workflows/deploy.yml#L124

.github/workflows/deploy.yml
```diff
@@ -24,6 +24,8 @@ jobs:
             default_max_vcpus: 600
             expanded_max_vcpus: 1600
             required_surplus: 1000
+            permission_boundary_arn: ''
+            deploy_lamdas_in_vpc: false
```
asjohnston-asf, line 28:
```suggestion
            permissions_boundary_arn: ''
            deploy_lambdas_in_vpc: false
```

.github/actions/deploy-hyp3/action.yml
```diff
@@ -56,6 +56,12 @@ inputs:
   REQUIRED_SURPLUS:
     description: "Amount by which month-to-date budget must exceed month-to-date spending to increase fleet size, in dollars. Ignored when DefaultMaxvCpus = ExpandedMaxvCpus."
     required: true
+  PERMISSIONS_BOUNDARY_ARN:
+    description: "Arn of permissions boundary to provide to any created IAM roles"
+    required: true
+  DEPLOY_LAMBDAS_IN_VPC:
+    description: "Whether or not to deploy lambda functions in a VPC"
+    required: false
```
asjohnston-asf, line 64: The stack expects the string `true` or `false`, so if we're not requiring this input we should at least default to the string `false`

.github/actions/deploy-hyp3/action.yml
```diff
@@ -97,4 +103,6 @@ runs:
                 DefaultMaxvCpus='${{ inputs.DEFAULT_MAX_VCPUS }}' \
                 ExpandedMaxvCpus='${{ inputs.EXPANDED_MAX_VCPUS }}' \
                 MonthlyComputeBudget='${{ inputs.MONTHLY_COMPUTE_BUDGET }}' \
-                RequiredSurplus='${{ inputs.REQUIRED_SURPLUS }}'
+                RequiredSurplus='${{ inputs.REQUIRED_SURPLUS }}' \
+                PermissionsBoundaryArn='${{ inputs.PERMISSIONS_BOUNDARY_ARN }}' \
+                DeployLambdasInVpc='${{ inputes.DEPLOY_LAMBDAS_IN_VPC }}'
```
asjohnston-asf, line 108:
```suggestion
                DeployLambdasInVpc='${{ inputs.DEPLOY_LAMBDAS_IN_VPC }}'
```

.github/workflows/deploy.yml
asjohnston-asf: Both the the matrix field names are typoed. They're consistently typoed, so it'll work, but I'd prefer to see them fixed throughout.

asjohnston-asf (approved):

PR #732, opened 2021-12-08
allow secret to default to empty string

Jlrine2:


asjohnston-asf (approved):

asjohnston-asf:
looks like the secret was never actually created for the UAT environment?

PR #736, opened 2021-12-09
skeleton of converting DeployLambasInVpc stack param to EarthdataClou…

asjohnston-asf:
…d to also manage bucket policy

### DIFF
--- a/.github/actions/deploy-hyp3/action.yml
+++ b/.github/actions/deploy-hyp3/action.yml
@@ -60,7 +60,7 @@ inputs:
     description: "Arn of permissions boundary to provide to any created IAM roles"
     required: true
   EARTHDATA_CLOUD:
-    description: "Whether or not to exclude resources/configurations incompatable with Earthdata Cloud"
+    description: "Modify resources/configurations to be compatible with Earthdata Cloud"
     required: true
 
 runs:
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -107,7 +107,7 @@ jobs:
             default_max_vcpus: 1600
             expanded_max_vcpus: 1600
             required_surplus: 0
-            deploy_lambdas_in_vpc: false
+            earthdata_cloudv: false
```
asjohnston-asf, line 110:
```suggestion
            earthdata_cloud: false
```

.github/actions/deploy-hyp3/action.yml
```diff
@@ -59,8 +59,8 @@ inputs:
   PERMISSIONS_BOUNDARY_ARN:
     description: "Arn of permissions boundary to provide to any created IAM roles"
     required: true
-  DEPLOY_LAMBDAS_IN_VPC:
-    description: "Whether or not to deploy lambda functions in a VPC"
+  EARTHDATA_CLOUD:
+    description: "Whether or not to exclude resources/configurations incompatable with Earthdata Cloud"
```
asjohnston-asf, line 63: We're excluding the bucket policy, but deploying the lambdas into the VPC feels like adding rather than removing. Maybe
```suggestion
    description: "Modify resources/configurations to be compatible with Earthdata Cloud"
```

Jlrine2 (approved):

PR #780, opened 2022-01-15
Grant CloudFront access to S3 content for Earthdata Cloud deployments

asjohnston-asf:
TODO: add ORIGIN_ACCESS_IDENTITY_ID secret to hyp3-edc-uat and hyp3-edc-prod environments

Adds an optional `OriginAccessIdentityId` stack parameter used to grant CloudFront read access to the S3 content bucket when deploying into Earthdata Cloud.

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html

This will allow data to be accessed via CloudFront URLs, but additional changes will still be needed to have the API return CloudFront URLs rather than S3 URLs.


jhkennedy (approved):

apps/main-cf.yml.j2
```diff
@@ -284,7 +289,7 @@ Resources:
         Rules:
           - ObjectOwnership: BucketOwnerEnforced
 
-  {% if security_environment in ('ASF', 'JPL-public') %}
+  {% if security_environment != 'JPL' %}
```
asjohnston-asf, line 292: I dislike having this condition here. I'd love to brainstorm a no-op `Principal` to use for `JPL` environments in the lower if/else blocks instead

jtherrmann:
Added `ORIGIN_ACCESS_IDENTITY_ID` secret to uat and prod environments.

jtherrmann (approved):

asjohnston-asf:
To validate this change:

Here's an underlying object in S3 (only accessible with AWS credentials):
s3://hyp3-edc-uat-contentbucket-89wtw9wpx9w7/0021798f-aec8-4de0-b50d-5c080f1d31ad/S1A_IW_20210818T081819_SHP_RTC30_G_gaunem_8335.png

Here's the direct S3 URL that would currently be returned by the API. It should result in 403 Access Denied both before and after this PR.
https://hyp3-edc-uat-contentbucket-89wtw9wpx9w7.s3.us-west-2.amazonaws.com/0021798f-aec8-4de0-b50d-5c080f1d31ad/S1A_IW_20210818T081819_SHP_RTC30_G_gaunem_8335.png

Here's the corresponding URL exposing that object via CloudFront. It currently gives 403 Forbidden, but should successfully serve the file after this PR.
https://d1riv60tezqha9.cloudfront.net/0021798f-aec8-4de0-b50d-5c080f1d31ad/S1A_IW_20210818T081819_SHP_RTC30_G_gaunem_8335.png

Here's a url for a non-existent file. It currently gives 403 Forbidden, but should give 404 Not Found after this PR.
https://d1riv60tezqha9.cloudfront.net/foo/bar.png

PR #791, opened 2022-01-21
Add JPL Deployment

jhkennedy:
I've successfully deployed to JPL from this branch


apps/api/api-cf.yml
```diff
@@ -97,8 +102,10 @@ Resources:
       Stage: $default
 
   ApiRole:
-    Type: AWS::IAM::Role
+    Type: !If [UseJPLRoles, Custom::JplRole, AWS::IAM::Role]
```
jhkennedy, line 105: It appears using conditionals for `Type:`  is very much not supported, and you end up with:
```
Template format error: [/Resources/Role] Every Type member must be a string.
```

which cfn-lint correctly warns about:
```
E3001 Type has to be a string at 
```

My guess is we'll have to do something like:
* write two roles: `ApiRole` and `JplApiRole`
* conditionally deploy those roles
* conditionally reference those roles

asjohnston-asf: I'm considering using jinja to inject the `Type`, since I'm worried about copy/paste errors in the `ApiRole` vs `JplApiRole` approach. That would effectively make *every* cloudformation template a jinja template, though, and doing through static analysis would mean rendering/validate for both platforms.

jhkennedy: I'm about 1/4 the way through doing my dual role suggestion above, and I think I agree we should jinja-ize it. This is *a lot* of copy-paste just to implement a if condition for `Type`. 

We already render before static analysis, so I think that's okay -- we can do a matrix of the security environments for the static analysis. 

It'll probably be more flexible in the long run (David mentioned deploying to CalTech as well today) but oof, this is all much bigger than I want though.

Makefile
```diff
@@ -40,15 +41,13 @@ openapi-validate: render
 	prance validate --backend=openapi-spec-validator apps/api/src/hyp3_api/api-spec/openapi-spec.yml
 
 cfn-lint: render
-	cfn-lint --info --ignore-checks W3002 --template `find . -name *-cf.yml`
+	cfn-lint --info -i W3002 -i E3008 --template `find . -name *-cf.yml`
```
jhkennedy, line 44: For JPL we get a
```
E3008 Property "Roles" can Ref to resources of types [AWS::IAM::Role] at Resources/InstanceProfile/Properties/Roles/0/Ref
./apps/compute-cf.yml:166:7
```

asjohnston-asf, line 44: I prefer the long-form parameter names for readability, and it looks like there's no need to repeat the parameter flag for each entry:
```suggestion
	cfn-lint --info --ignore-checks W3002 E3008 --template `find . -name *-cf.yml`
```

apps/render_cf.py
```diff
@@ -11,21 +11,7 @@ def snake_to_pascal_case(input_string: str):
...
-        f.write(template.render(job_types=job_types, json=json, snake_to_pascal_case=snake_to_pascal_case))
-
-
-def render_templates(job_types):
-    env = get_env()
-    for template_file in Path('.').glob('**/*.j2'):
-        render_template(PurePosixPath(template_file), job_types, env)
-
-
-def get_env():
```
jhkennedy, line 28: I found a lot of this unnecessary and the logic easier to follow with just a `render_templates` fn. I can back out some or all of it though.

asjohnston-asf (approved):

PR #800, opened 2022-01-26
Bug sub timeout

jacquelynsmale:


asjohnston-asf:
Already discussed with @jacquelynsmale but posting here for the group:

We disable a subscription once it's end date plus a grace period has passed: https://github.com/ASFHyP3/hyp3/blob/develop/apps/process-new-granules/src/process_new_granules.py#L79

That logic was based on the assumption that `handle_subscription` would always process all outstanding granules for the subscription before it's disabled. With this PR, that's no longer true, since `handle_subscription` will only process up to 20 outstanding granules.

We've proposed adding an additional `and len(get_unprocessed_granules(subscription)) == 0` check prior to disabling the subscription.

asjohnston-asf (approved):

PR #829, opened 2022-02-08
Require HttpTokens in metadata for IMDSv2

jacquelynsmale:

### DIFF
--- a/apps/compute-cf.yml.j2
+++ b/apps/compute-cf.yml.j2
@@ -51,6 +51,8 @@ Resources:
     Type: AWS::EC2::LaunchTemplate
     Properties:
       LaunchTemplateData:
+        MetadataOptions:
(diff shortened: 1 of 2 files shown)


apps/compute-cf.yml.j2
```diff
@@ -51,6 +51,8 @@ Resources:
     Type: AWS::EC2::LaunchTemplate
     Properties:
       LaunchTemplateData:
+        MetadataOptions:
+            HttpTokens: 'required'
```
asjohnston-asf, line 55: We're using two-space indentation for yaml files, although I think this is syntactically valid either way. And I lean towards no using quotes for strings in yaml unless specifically required.
```suggestion
          HttpTokens: required
```

asjohnston-asf (approved):

PR #836, opened 2022-02-15
Record job processing time in database

jtherrmann:

### DIFF
--- a/apps/check-processing-time/check-processing-time-cf.yml.j2
+++ b/apps/check-processing-time/check-processing-time-cf.yml.j2
@@ -1,0 +1,84 @@
+AWSTemplateFormatVersion: 2010-09-09
+
+Parameters:
+
+  PermissionsBoundaryPolicyArn:
+    Type: String
+    Default: ""
+
+  SecurityGroupId:
+    Type: String
+
+  SubnetIds:
+    Type: CommaDelimitedList
+
+Conditions:
+
+  UsePermissionsBoundary: !Not [!Equals [!Ref PermissionsBoundaryPolicyArn, ""]]
+
+  LambdasInVpc: !Not [!Equals [!Ref SecurityGroupId, ""]]
+
+Outputs:
+
+  LambdaArn:
+    Value: !GetAtt Lambda.Arn
+
+Resources:
+
+  LogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: !Sub "/aws/lambda/${Lambda}"
+      RetentionInDays: 90
+
+  Role:
+    Type: {{ 'Custom::JplRole' if security_environment == 'JPL' else 'AWS::IAM::Role' }}
+    Properties:
+      {% if security_environment == 'JPL' %}
+      ServiceToken: !ImportValue Custom::JplRole::ServiceToken
+      Path: /account-managed/hyp3/
+      {% endif %}
+      AssumeRolePolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          Action: sts:AssumeRole
+          Principal:
+            Service: lambda.amazonaws.com
+          Effect: Allow
+      PermissionsBoundary: !If [UsePermissionsBoundary, !Ref PermissionsBoundaryPolicyArn, !Ref AWS::NoValue]
+      ManagedPolicyArns:
+        - !If [LambdasInVpc, arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole, !Ref AWS::NoValue]
+        - !Ref Policy
+
+  Policy:
+    Type: {{ 'Custom::JplPolicy' if security_environment == 'JPL' else 'AWS::IAM::ManagedPolicy' }}
+    Properties:
+      {% if security_environment == 'JPL' %}
+      ServiceToken: !ImportValue Custom::JplPolicy::ServiceToken
+      Path: /account-managed/hyp3/
+      {% endif %}
+      PolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: Allow
+            Action:
+              - logs:CreateLogStream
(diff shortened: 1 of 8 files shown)


asjohnston-asf:
We'll need to give the Step Function IAM role permission to invoke the new lambda function at https://github.com/ASFHyP3/hyp3/blob/processing-time/apps/workflow-cf.yml.j2#L133

apps/check-processing-time/check-processing-time-cf.yml.j2
```diff
@@ -0,0 +1,84 @@
+AWSTemplateFormatVersion: 2010-09-09
+
+Parameters:
+
+  PermissionsBoundaryPolicyArn:
+    Type: String
+    Default: ""
```
asjohnston-asf, line 7: We've avoided setting default parameters anywhere outside of `main.cf.j2`. In the past it's helped catch errors where the parent stack wasn't correctly passing a value to a child stack.
```suggestion
```

apps/step-function.json.j2
```diff
@@ -1,6 +1,14 @@
 {
-  "StartAt": "JOB_RUNNING",
+  "StartAt": "SET_DEFAULT_RESULTS",
   "States": {
+    "SET_DEFAULT_RESULTS": {
```
asjohnston-asf, line 4: Not blocking, but I think I'd prefer `JOB_RUNNING` still be the first task, since we want to update the status code in the database from PENDING to RUNNING as quickly as possible. No real risk in this case, but putting other tasks in front of JOB_RUNNING might cause us to lose sight of it down the road.

apps/check-processing-time/src/check_processing_time.py
```diff
@@ -0,0 +1,14 @@
...
+    attempts.sort(key=lambda attempt: attempt['StartedAt'])
+    final_attempt = attempts[-1]
+    return final_attempt['StoppedAt'] - final_attempt['StartedAt']
+
+
+def lambda_handler(event, context):
+    results = event['processing_results']
+    if 'Attempts' in results:
+        return get_time_from_attempts(results['Attempts'])
+    return get_time_from_attempts(json.loads(results['Cause'])['Attempts'])
```
asjohnston-asf, line 14: Not blocking, but could separate the parsing of attempts from the processing to make this a bit more expressive:
```suggestion
    if 'Attempts' in results:
        attempts = results['Attempts']
    else:
        attempts = json.loads(results['Cause'])['Attempts']

    return get_time_from_attempts(attempts)
```

apps/check-processing-time/src/check_processing_time.py
```diff
@@ -0,0 +1,14 @@
+import json
+
+
+def get_time_from_attempts(attempts):
+    attempts.sort(key=lambda attempt: attempt['StartedAt'])
+    final_attempt = attempts[-1]
+    return final_attempt['StoppedAt'] - final_attempt['StartedAt']
```
asjohnston-asf, line 7: I think `StartedAt` and `StoppedAt` are in milliseconds, so we may need to divide by 1,000 here
https://docs.aws.amazon.com/batch/latest/APIReference/API_AttemptDetail.html#Batch-Type-AttemptDetail-startedAt

tests/test_check_processing_time.py
```diff
@@ -0,0 +1,54 @@
+import check_processing_time
+
+
+def test_single_attempt():
+    attempts = [{'Container': {}, 'StartedAt': 1644423555778, 'StatusReason': '', 'StoppedAt': 1644425055200}]
+    result = check_processing_time.get_time_from_attempts(attempts)
+    assert result == 1644425055200 - 1644423555778
```
asjohnston-asf, line 7: Test coverage looks good. In general, I prefer a test's final `assert` to compare against a constant to minimize the amount of thinking a reader has to do. Maybe one test should include real-ish timestamps, but I suspect most of them can be simplified to something like:
```suggestion
    attempts = [{'Container': {}, 'StartedAt': 11, 'StatusReason': '', 'StoppedAt': 13}]
    result = check_processing_time.get_time_from_attempts(attempts)
    assert result == 2
```

apps/check-processing-time/check-processing-time-cf.yml.j2
jtherrmann: Makes sense, I also removed that line from the `get-files` template (which is where I originally copied it from).

apps/check-processing-time/src/check_processing_time.py
jtherrmann: Ah, I meant to double-check those units. Thanks for the link! I was searching for that documentation but couldn't find it.

jtherrmann: I decided not to do any rounding because it could result in failed jobs with < 500 ms runtimes being shown as having runtime `0`, which should indicate that no processing was attempted. But I could instead always round up, if we'd rather `processing_time_in_seconds` be an integer instead of a float.

apps/step-function.json.j2
jtherrmann: I could easily make `SET_DEFAULT_RESULTS` be the second state if it only needs to set one field in `$.results`, but if it ever needs to set multiple fields, then I'm not sure how to do that without overwriting the `results` object, given the [documentation](https://states-language.net/spec.html#filters):

> If the raw input has a field at the location addressed by the ResultPath value then in the output that field is discarded and overwritten by the state's result.

So it seems necessary for `SET_DEFAULT_RESULTS` to create `$.results` before `JOB_RUNNING` creates `$.results.job_running`. Unless I'm misunderstanding how `ResultPath` works and it would actually just create the missing fields without overwriting the entire object?

asjohnston-asf: That checks out, no changes necessary, then :+1:

asjohnston-asf (approved):

PR #839, opened 2022-02-15
convert float values to Decimals in dynamo.jobs.update_job

asjohnston-asf:
DynamoDB doesn't support the `float` data type in Python; any data values of type `Number` in dynamodb should be of type `Decimal` in python.

We correctly convert any float values to decimals before inserting them into the database in dynamo.jobs.put_jobs: https://github.com/ASFHyP3/hyp3/blob/dbd9c7703525ed1c01faa4e3fc51055ed7fe3c05/lib/dynamo/dynamo/jobs.py#L52

but we'd missed doing the necessary conversion in dynamo.jobs.update_job.

### DIFF
--- a/tests/test_dynamo/test_jobs.py
+++ b/tests/test_dynamo/test_jobs.py
@@ -426,7 +426,7 @@ def test_update_job(tables):
     for item in table_items:
         tables.jobs_table.put_item(Item=item)
 
-    job = {'job_id': 'job1', 'status_code': 'status2'}
+    job = {'job_id': 'job1', 'status_code': 'status2', 'processing_time_in_seconds': 1.23}
     dynamo.jobs.update_job(job)
 
     response = tables.jobs_table.scan()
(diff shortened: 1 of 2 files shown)


jhkennedy (approved):

tests/test_dynamo/test_jobs.py
```diff
@@ -437,6 +437,7 @@ def test_update_job(tables):
             'user_id': 'user1',
             'status_code': 'status2',
             'request_time': '2000-01-01T00:00:00+00:00',
+            'processing_time_in_seconds': Decimal('1.23'),
```
jtherrmann, line 440: Are we concerned about the fact that using `Decimal` results in capturing the entire float value in the database, i.e. `Decimal(1.23)` is equivalent to `Decimal('1.229999999999999982236431605997495353221893310546875')`? Would it be better to ensure that the "actual" decimal values get put in the database?

jtherrmann: Ah, my mistake. I did not notice that the float is converted to a string first.

jtherrmann (approved):

PR #860, opened 2022-02-23
Differentiate JPL security environments

jhkennedy:
JPL by default disallows public buckets in accounts via the account wide S3 Block All Public Access settings. With some effort, Block All Public Access can be disabled by the JPL Cloud team at the account level which allows deployment of public buckets. 

This 
* changes the `JPL` security environment to *not* add a public bucket policy to the HyP3 content bucket
* Adds a `JPL-public` secuirt environment that *will* add a public bucket policy to the HyP3 content bucket

### DIFF
--- a/apps/main-cf.yml.j2
+++ b/apps/main-cf.yml.j2
@@ -245,7 +245,7 @@ Resources:
             AllowedOrigins:
               - "*.asf.alaska.edu"
 
-  {% if security_environment != 'EDC' %}
+  {% if security_environment in ('ASF', 'JPL-public') %}
   BucketPolicy:
     Type: AWS::S3::BucketPolicy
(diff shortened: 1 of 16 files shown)


apps/main-cf.yml.j2
```diff
@@ -245,7 +245,7 @@ Resources:
             AllowedOrigins:
               - "*.asf.alaska.edu"
 
-  {% if security_environment != 'EDC' %}
+  {% if security_environment in ('ASF', 'JPL-public') %}
```
jhkennedy, line 248: I was able to deploy into the new `hyp3-a19-jpl` environment by removing this bucket policy

asjohnston-asf (approved):

PR #861, opened 2022-02-23
Custom instance sizes

jhkennedy:
* makes instance sizes available to the compute environment customizable at render
* makes the vCPU and Memory configuration a job_spec paremter

This should allow us to, for example, 
* deploy a custom GUNW stack with an optimized instance size
* finally bring the watermap deployment into our normal deployment worklfow

### DIFF
--- a/.github/workflows/deploy-daac.yml
+++ b/.github/workflows/deploy-daac.yml
@@ -26,6 +26,7 @@ jobs:
             required_surplus: 1000
             security_environment: ASF
             ami_id: /aws/service/ecs/optimized-ami/amazon-linux-2/recommended/image_id
+            instance_types: r5d.xlarge
 
           - environment: hyp3-test
             domain: hyp3-test-api.asf.alaska.edu
@@ -45,6 +46,7 @@ jobs:
             required_surplus: 1000
             security_environment: ASF
             ami_id: /aws/service/ecs/optimized-ami/amazon-linux-2/recommended/image_id
+            instance_types: r5d.xlarge
 
           - environment: hyp3-edc-prod
             domain: ''
@@ -59,6 +61,7 @@ jobs:
             required_surplus: 1000
             security_environment: EDC
             ami_id: image_id_ecs_amz2
+            instance_types: r5d.xlarge
 
           - environment: hyp3-edc-uat
             domain: ''
@@ -73,6 +76,7 @@ jobs:
             required_surplus: 1000
             security_environment: EDC
             ami_id: image_id_ecs_amz2
+            instance_types: r5d.xlarge
 
     environment:
       name: ${{ matrix.environment }}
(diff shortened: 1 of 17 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+## [2.12.0]
+### Added
+- Instance types available to the compute environment is configurable when rendering CloudFormation templates
+  via the `--instance-types` argument to [`render_cf.py`](apps/render_cf.py).
+
+### Changed
+- The `job_spec_files` positional argument to [`render_cf.py`](apps/render_cf.py) has been switched to a
+  required `--job-spec-files` optional argument to support multiple open-ended arguments.
+
+
```
asjohnston-asf, line 17:
```suggestion
```

.github/workflows/deploy-enterprise.yml
```diff
@@ -50,6 +52,7 @@ jobs:
             required_surplus: 0
             security_environment: ASF
             ami_id: /aws/service/ecs/optimized-ami/amazon-linux-2/recommended/image_id
+            instance_types: c5d.xlarge
```
jhkennedy, line 55: I've updated the GUNW deployments (this and next two) to use the cost-optimized instance type based on our optimization study.

job_spec/INSAR_ISCE.yml
```diff
@@ -41,3 +41,5 @@ INSAR_ISCE:
     - Ref::secondary_granules
   validators: []
   timeout: 10800
+  vcpu: 1
+  memory: 7600
```
jhkennedy, line 45: I'm not sure this is the correct # to use for the GUNW jobs?

.github/workflows/deploy-enterprise.yml
jhkennedy: Costs have remained stable for these machines the last 3 months:
image

job_spec/INSAR_ISCE.yml
jhkennedy:
```suggestion
  memory: 7500
```

.github/workflows/deploy-daac.yml
```diff
@@ -45,6 +46,9 @@ jobs:
             required_surplus: 1000
             security_environment: ASF
             ami_id: /aws/service/ecs/optimized-ami/amazon-linux-2/recommended/image_id
+            instance_types: >-
+              r5d.xlarge
+              r5dn.xlarge
```
jhkennedy, line 51: Utilizing the multiple instance types here and in the `hyp3-its-live` deployment. 

This will let us test it in test, but I'm not sure we'll be able to guarantee the new type will be picked up?

asjohnston-asf (approved):

PR #864, opened 2022-02-24
Refactor step function to consolidate autorift steps

jtherrmann:
[Visualization of the new workflow.](https://user-images.githubusercontent.com/19476938/155619588-5bc1f976-f628-4f17-a9b1-7de74d33fcda.png)

### DIFF
--- a/apps/step-function.json.j2
+++ b/apps/step-function.json.j2
@@ -1,6 +1,13 @@
+{# TODO reorder States list to reflect current workflow structure #}
 {
-  "StartAt": "SET_DEFAULT_RESULTS",
+  "StartAt": "SET_DEFAULT_CONTAINER_OVERRIDES",
   "States": {
+    "SET_DEFAULT_CONTAINER_OVERRIDES": {
+      "Type": "Pass",
+      "Result": {},
+      "ResultPath": "$.container_overrides",
+      "Next": "SET_DEFAULT_RESULTS"
+    },
     "SET_DEFAULT_RESULTS": {
       "Type": "Pass",
       "Result": {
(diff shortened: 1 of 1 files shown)


asjohnston-asf:
:thinking: I'm seeing issues when the template is rendered without the job_spec/AUTORIFT.yml job type, since the memory-related steps are created, but the AUTORIFT task is missing.

e.g. `make render files="job_spec/RTC_GAMMA.yml job_spec/INSAR_GAMMA.yml"`

Think this ties into the "finished refactoring how jobs are defined" TODO; I haven't thought of a quick fix yet.

jtherrmann:
> thinking I'm seeing issues when the template is rendered without the job_spec/AUTORIFT.yml job type, since the memory-related steps are created, but the AUTORIFT task is missing.
> 
> e.g. `make render files="job_spec/RTC_GAMMA.yml job_spec/INSAR_GAMMA.yml"`
> 
> Think this ties into the "finished refactoring how jobs are defined" TODO; I haven't thought of a quick fix yet.

Ah, I did not realize there was a situation in which the template is rendered without one of the job types. I think that the long-term solution is we will start defining jobs as step functions, rather than single batch steps. So we could just delay merging this PR until I convert jobs to step functions.

asjohnston-asf:
Adding `{% if 'AUTORIFT' in job_types %}` around the `INSPECT_GRANULE_TYPE/USE_SENTINEL2_MEMORY/USE_LANDSAT_MEMORY` task definitions looks like it does the trick, but that feels like it just makes the "temporary hack" bigger

jtherrmann:
> Adding `{% if 'AUTORIFT' in job_types %}` around the `INSPECT_GRANULE_TYPE/USE_SENTINEL2_MEMORY/USE_LANDSAT_MEMORY` task definitions looks like it does the trick, but that feels like it just makes the "temporary hack" bigger

Agreed, I may as well just proceed with doing the full refactor (i.e. converting jobs to step functions), if we're sure that's the final goal.

asjohnston-asf:
The `files` parameter in the two deploy actions lists what's deployed to each enviroment; the enterprise deployment have a variety of configurations
https://github.com/ASFHyP3/hyp3/blob/develop/.github/workflows/deploy-daac.yml
https://github.com/ASFHyP3/hyp3/blob/develop/.github/workflows/deploy-enterprise.yml

PR #916, opened 2022-03-22
Add an environmnet.yml

jhkennedy:
Simply specifies our python version and then `pip` installs all our requirements. makes it so our normal `conda env update -f environment.yml` also works with `hyp3`

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,4 +6,4 @@ dependencies:
   - python=3.9
   - pip
   - pip:
-      -r requirements-all.txt
+      - -r requirements-all.txt
--- a/environment.yml
+++ b/environment.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - nodefaults
 dependencies:
-  - python=3.8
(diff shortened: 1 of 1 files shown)


jhkennedy:
Bah, doesn't work as expected as conda creates a separate pip requirements file... 

```
Installing pip dependencies: / Ran pip subprocess with arguments:
['/home/jhkennedy/mambaforge/envs/hyp3/bin/python', '-m', 'pip', 'install', '-U', '-r', '/home/jhkennedy/Documents/Code/HyP3/hyp3/condaenv.imj2n6jd.requirements.txt']
```

jhkennedy:
Okay, after exporting [`CONDA_TEST_SAVE_TEMPS`](https://github.com/conda/conda/blob/master/conda_env/installers/pip.py#L58) and seeing what was happening, it turns out I'm a dummy and didn't write the YAML file correctly. 

This should work now.

asjohnston-asf (approved):

PR #924, opened 2022-03-23
Make lifecycle policy optional

jtherrmann:


jhkennedy (commented):
Overall looks good to me!

Some thoughts:
* it's probably worth adding that parameter to the `cfn-lint` matrix here: https://github.com/ASFHyP3/hyp3/blob/develop/.github/workflows/static-analysis.yml#L23-L24
  ```
          matrix:
          security_environment: [ASF, EDC, JPL, JPL-public]
          lifecycle_policy: [yes, no]
  ```
  and pass it to the make command here: https://github.com/ASFHyP3/hyp3/blob/develop/.github/workflows/static-analysis.yml#L35

* That'll likely cause the cfn-lint to complain about the unused `ProductLifetimeInDays` CF parameter when `lifecycle_policy=no`.  So you could also remove this with jinja: https://github.com/ASFHyP3/hyp3/blob/develop/apps/main-cf.yml.j2#L24-L28
  
  which would mean needing to adjust the cloudformation deploy argument to be optional like:
  https://github.com/ASFHyP3/hyp3/blob/develop/.github/actions/deploy-hyp3/action.yml#L90-L91

  and setting `product_lifetime_in_days` in the matrix to an empty string for AVO

* In the deploy matrices, you may want to put `lifecycle_policy` and `product_lifetime_in_days` next to each other since they are used

jhkennedy (approved):
LGTM!

jtherrmann:
After some discussion, we realized this issue would require more work (e.g. updating `get_files` to return `None` for expiration time when appropriate and inserting a null value into the dynamodb table), so we decided to just set an extremely long expiration time when we want products to last forever.

PR #954, opened 2022-04-07
Add flood maps

jacquelynsmale:


job_spec/FLOOD_MAP.yml
```diff
@@ -0,0 +1,69 @@
+WATER_MAP:
```
jtherrmann, line 1:
```suggestion
FLOOD_MAP:
```

job_spec/FLOOD_MAP_10M.yml
```diff
@@ -0,0 +1,69 @@
+WATER_MAP:
```
jtherrmann, line 1:
```suggestion
FLOOD_MAP:
```

jtherrmann (changes requested):
~~I'm not seeing a difference between `FLOOD_MAP.yml` and `FLOOD_MAP_10M.yml`, but I assume there should be? You might try running a diff of `WATER_MAP.yml` and `WATER_MAP_10M.yml` to see what might need to change (it's only a few lines), though I'm not sure if the changes would be exactly the same for the flood map spec.~~

jtherrmann (commented):
Looks good apart from the requested changes. I don't have the background knowledge to verify that all of the parameters, default values, etc. are correct, so I'll let you judge if you're confident in those or if you want someone else with more expertise to add their review.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.14.5]
```
jtherrmann, line 7: This feels like a minor change, rather than a patch, to me.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
...
+### Added
+- Flood depth map option added to test.
```
jtherrmann, line 9: Just for added clarity, I'd suggest something along the lines of:

```suggestion
- Added `FLOOD_MAP` job spec (currently limited to `hyp3-test`).
```

.github/workflows/deploy-daac.yml
```diff
@@ -42,6 +42,7 @@ jobs:
               job_spec/RTC_GAMMA.yml
               job_spec/INSAR_ISCE.yml
               job_spec/WATER_MAP.yml
+              job_spec/FLOOD_MAP.yml
```
jtherrmann, line 45: No longer needed:
```suggestion
```

jtherrmann (changes requested):

jtherrmann:
Self-reminder to diff the two job spec files before final approval.

job_spec/WATER_MAP.yml
```diff
@@ -62,6 +62,33 @@ WATER_MAP:
         description: The average membership to the fuzzy indicators required for a water pixel
         default: 0.45
         type: number
+    include_flood_depth:
+      api_schema:
+        description: Produce a flood depth map from water products.
+        default: false
+        type: boolean
+    estimator:
```
jtherrmann, line 70: Just as an example:
```suggestion
    flood_depth_estimator:
```

job_spec/WATER_MAP.yml
```diff
@@ -83,6 +110,16 @@ WATER_MAP:
     - Ref::hand_fraction
     - --membership-threshold
     - Ref::membership_threshold
+    - --include_flood_depth
+    - Ref::include_flood_depth
+    - --estimator
+    - Ref::estimator
```
jtherrmann, line 116: Continued from above:
```suggestion
    - Ref::flood_depth_estimator
```

job_spec/WATER_MAP.yml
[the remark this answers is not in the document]
jtherrmann: @asjohnston-asf Do you know of a better way to make it clear that these parameters should only be used if `include_flood_depth` is `true`?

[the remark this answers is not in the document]
asjohnston-asf: I can't think of anything better. Your suggestion for the parameter name(s) is a good one. Could also call out "ignored when include_flood_depth is false" in the description for each.

Could potentially combine `include_flood_depth` with another parameter, say add `none` as a `flood_depth_estimator` option, but that still leaves the problem of linking with the other parameters.

Could address the relationship at the API level by having the flood depth parameters be their own dictionary, e.g. 
```
"job_parameters": {
  "resolution": 30.0,
  "flood_depth": {
    "estimator": "nmad",
    "water_level_sigma": 3.0
  }
}
```
but I suspect that would take significant refactoring to get those values handed off as individual command line parameter key/value pairs to the Batch container

jtherrmann (approved):

PR #986, opened 2022-04-29
Don't change status from PENDING to RUNNING until Batch job starts running

jtherrmann:
Summary:
* Add a new field `execution_started` to the jobs table that represents whether the step function has been executed for a given job.
* Only start executions for jobs with status `PENDING` and whose `execution_started` field is false or does not exist (to account for pending jobs that were added before this feature was merged).
* Update hyp3 job status to `RUNNING` when the Batch job starts running.

TODO:
- [x] Should we add the `execution_started` field to the API spec?

### DIFF
--- a/apps/handle-batch-event/handle-batch-event-cf.yml.j2
+++ b/apps/handle-batch-event/handle-batch-event-cf.yml.j2
@@ -5,6 +5,9 @@ Parameters:
   JobQueueArn:
     Type: String
 
+  JobsTable:
+    Type: String
+
   {% if security_environment == 'EDC' %}
   PermissionsBoundaryPolicyArn:
     Type: String
@@ -60,6 +63,9 @@ Resources:
               - logs:CreateLogStream
               - logs:PutLogEvents
             Resource: !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/*"
+          - Effect: Allow
+            Action: dynamodb:UpdateItem
+            Resource: !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${JobsTable}*"
 
   Lambda:
     Type: AWS::Lambda::Function
--- a/apps/handle-batch-event/handle-batch-event-cf.yml.j2
+++ b/apps/handle-batch-event/handle-batch-event-cf.yml.j2
@@ -91,4 +91,5 @@ Resources:
           status:
             - RUNNING
       Targets:
-        - !GetAtt Lambda.Arn
+        - Arn: !GetAtt Lambda.Arn
+          Id: lambda-target
--- a/apps/handle-batch-event/handle-batch-event-cf.yml.j2
+++ b/apps/handle-batch-event/handle-batch-event-cf.yml.j2
@@ -70,6 +70,9 @@ Resources:
   Lambda:
     Type: AWS::Lambda::Function
     Properties:
+      Environment:
+        Variables:
+          JOBS_TABLE_NAME: !Ref JobsTable
       Code: src/
       Handler: handle_batch_event.lambda_handler
       MemorySize: 128
--- a/apps/handle-batch-event/handle-batch-event-cf.yml.j2
+++ b/apps/handle-batch-event/handle-batch-event-cf.yml.j2
@@ -102,3 +102,11 @@ Resources:
       Targets:
         - Arn: !GetAtt Lambda.Arn
           Id: lambda-target
+
+  PermissionForEventsToInvokeLambda:
+    Type: AWS::Lambda::Permission
+    Properties:
+      FunctionName: !GetAtt Lambda.Arn
+      Action: lambda:InvokeFunction
+      Principal: events.amazonaws.com
+      SourceArn: !GetAtt EventRule.Arn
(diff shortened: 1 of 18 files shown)


apps/handle-batch-event/handle-batch-event-cf.yml.j2
```diff
@@ -0,0 +1,95 @@
...
+      Path: /account-managed/hyp3/
+      {% endif %}
+      PolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: Allow
+            Action:
+              - logs:CreateLogStream
+              - logs:PutLogEvents
+            Resource: !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/*"
```
asjohnston-asf, line 62: We'll need additional permissions where for whatever `boto3` calls the Lambda will make:
`dynamodbUpdateItem` on the Jobs table

apps/handle-batch-event/handle-batch-event-cf.yml.j2
```diff
@@ -0,0 +1,95 @@
...
+
+  Lambda:
+    Type: AWS::Lambda::Function
+    Properties:
+      Code: src/
+      Handler: handle_batch_event.lambda_handler
+      MemorySize: 128
+      Role: !GetAtt Role.Arn
+      Runtime: python3.9
+      Timeout: 30
```
asjohnston-asf, line 72: we'll need a `JOBS_TABLE_NAME` environment variable for the `dynamo.update_job` function

apps/handle-batch-event/handle-batch-event-cf.yml.j2
```diff
@@ -0,0 +1,95 @@
+AWSTemplateFormatVersion: 2010-09-09
+
+Parameters:
+
+  JobQueueArn:
+    Type: String
+
```
asjohnston-asf, line 7: add a `JobsTableName` parameter to pass to the lambda function

apps/handle-batch-event/handle-batch-event-cf.yml.j2
```diff
@@ -0,0 +1,95 @@
...
+        detail-type:
+          - Batch Job State Change
+        detail:
+          jobQueue:
+            - !Ref JobQueueArn
+          status:
+            - RUNNING
+      Targets:
+        - Arn: !GetAtt Lambda.Arn
+          Id: lambda-target
```
asjohnston-asf, line 95: I haven't worked with EventBridge before, but I suspect we'll need a `AWS::Lambda::Permission` resource to give eventbridge permission to invoke the lambda function, similar to how we grant cloudwatch rules permission to invoke our scheduled lambdas, e.g. https://github.com/ASFHyP3/hyp3/blob/develop/apps/start-execution/start-execution-cf.yml.j2#L108

apps/main-cf.yml.j2
```diff
@@ -191,6 +191,18 @@ Resources:
         {% endif %}
       TemplateURL: process-new-granules/process-new-granules-cf.yml
 
+  HandleBatchEvent:
+    Type: AWS::CloudFormation::Stack
+    Properties:
+      Parameters:
+        JobQueueArn: !GetAtt Cluster.Outputs.JobQueueArn
+        {% if security_environment == 'EDC' %}
```
asjohnston-asf, line 199: we'll need to pass along the Jobs Table name

lib/dynamo/dynamo/jobs.py
```diff
@@ -136,3 +137,15 @@ def get_jobs_by_status_code(status_code: str, limit: int) -> List[dict]:
...
+
+
+def get_jobs_waiting_for_execution(limit: int) -> list[dict]:
+    table = DYNAMODB_RESOURCE.Table(environ['JOBS_TABLE_NAME'])
+    response = table.query(
+        IndexName='status_code',
+        KeyConditionExpression=Key('status_code').eq('PENDING'),
+        FilterExpression=Attr('execution_started').ne(True),
+        Limit=limit,
+    )
```
asjohnston-asf, line 149: we could call `get_jobs_by_status('PENDING')` here, and then filter by `execution_started` manually, but I'm not sure if that's an improvement or not

apps/handle-batch-event/handle-batch-event-cf.yml.j2
jtherrmann: Ah, yep, it's [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-rule.html#aws-resource-events-rule--examples) under "Invoke Lambda Function in Response to an Event".

lib/dynamo/dynamo/jobs.py
jtherrmann: I considered that, but didn't like it much. Although, I'm not sure `get_jobs_by_status_code` has any other usages now, so if there's no longer a need for it, we could just remove it.

asjohnston-asf: Agreed, I don't see any other calls to `get_jobs_by_status_code`, so I'd be in favor of eliminating that function and it's tests.

asjohnston-asf:
`handle_batch_event` depends on the `dynamo` library, so we'll need to:
- add `requirements-apps-handle-batch-event.txt` that includes `./lib/dynamo/`
- update `requirements-all.txt`
- update the `build` target in `Makefile`

asjohnston-asf (approved):

PR #993, opened 2022-05-03
Add default `get_files` values to step functions results

jtherrmann:
TODO:
- [x] Document expiration time as nullable in API spec.
- [x] Double-check how this will work in the SDK.


apps/step-function.json.j2
```diff
@@ -4,7 +4,11 @@
     "SET_DEFAULT_RESULTS": {
       "Type": "Pass",
       "Result": {
-        "processing_time_in_seconds": 0
+        "processing_time_in_seconds": 0,
+        "get_files": {
+          "logs": [],
+          "expiration_time": null
```
asjohnston-asf, line 10: Not sure what a good default `expiration_time` should be. I want to take a closer look at what assumptions `hyp3_sdk` makes about that field.

jtherrmann: Looks like it will have the same behavior whether the `expiration_time` field does not exist or is `None`. I can write additional tests for `hyp3-sdk` to cover the case where it is `None`.

jtherrmann:
Closed in favor of https://github.com/ASFHyP3/hyp3/pull/998

asjohnston-asf:
I think we should consider changing https://github.com/ASFHyP3/hyp3/blob/develop/apps/step-function.json.j2#L189 to put the results of get_files in a separate `get_files_error` field so that any errors don't clobber the default values? the same as we did for processing_time_in_seconds at https://github.com/ASFHyP3/hyp3/blob/develop/apps/step-function.json.j2#L215 ?

asjohnston-asf (approved):

PR #1010, opened 2022-05-12
Allow unlimited quota, priority override

jtherrmann:
TODO:

- [x] Add/update tests when happy with the intended behavior


lib/dynamo/dynamo/user.py
```diff
@@ -9,10 +10,21 @@ def get_user(user):
...
+        priority = user.get('priority')
+    else:
+        priority = None
+    return priority
+
+
+def get_max_jobs_per_month(user) -> Optional[int]:
+    user = get_user(user)
+    if user:
+        max_jobs_per_month = user['max_jobs_per_month']
```
asjohnston-asf, line 25: catch the case where a user has a record, but that record has no `max_jobs_per_month` field
```suggestion
        max_jobs_per_month = user.get('max_jobs_per_month')
```

asjohnston-asf: Hrmm, it's actually more complicated than that, since we'd want to return the default MONTHLY_JOB_QUOTA_PER_USER in this case, rather than None

lib/dynamo/dynamo/jobs.py
```diff
@@ -20,31 +20,46 @@ def _get_job_count_for_month(user) -> int:
...
+        remaining_jobs = None
+
     return max_jobs, previous_jobs, remaining_jobs
 
 
 def put_jobs(user_id: str, jobs: List[dict], fail_when_over_quota=True) -> List[dict]:
     table = DYNAMODB_RESOURCE.Table(environ['JOBS_TABLE_NAME'])
     request_time = format_time(datetime.now(timezone.utc))
 
+    priority_override = get_priority(user_id)
```
asjohnston-asf, line 40: I'd wait to declare this variable until immediately before its' used (around line 47), i.e. do all the `max_jobs` business logic, then do all the `priority` business logic

lib/dynamo/dynamo/jobs.py
```diff
@@ -20,31 +20,46 @@ def _get_job_count_for_month(user) -> int:
...
         if fail_when_over_quota:
             raise QuotaError(f'Your monthly quota is {max_jobs} jobs. You have {remaining_jobs} jobs remaining.')
         jobs = jobs[:remaining_jobs]
 
+    if priority_override is not None:
+        priority = lambda _: priority_override  # noqa: E731
+    elif max_jobs is None:
+        priority = lambda _: 0  # noqa: E731
+    else:
+        priority = lambda job_index: max(9999 - previous_jobs - job_index, 0)  # noqa: E731
```
asjohnston-asf, line 53: I like the approach, but I'm curious if there's a way to move this chunk of logic into a separate function to hide the details?

asjohnston-asf:
Should also add `nullable: true` for `max_jobs_per_month` and `remaining` in the OpenAPI specification at https://github.com/ASFHyP3/hyp3/blob/develop/apps/api/src/hyp3_api/api-spec/openapi-spec.yml#L302

https://swagger.io/docs/specification/data-models/data-types/#null

lib/dynamo/dynamo/jobs.py
asjohnston-asf: Also wondering whether the if/elif/else should specifically look for `previous_jobs is not None`, since that's the value being referenced in the `else` block?

jtherrmann: Here's an example of using a separate function and checking `previous_jobs` rather than `max_jobs`:

```python3
def _get_job_priority(priority_override: Optional[int], previous_jobs: Optional[int]):
    if priority_override is not None:
        priority = lambda _: priority_override
    elif previous_jobs is None:
        priority = lambda _: 0
    else:
        priority = lambda job_index: max(9999 - previous_jobs - job_index, 0)
    return priority
```
It could of course be refactored to instead check `previous_jobs is not None`. But regardless, I feel like this approach makes it more difficult to understand that we default to priority `0` when the user has no quota. I think checking `max_jobs is None` makes it more clear; but it seems a bit silly to pass `max_jobs` into the helper function just to check if it's `None`. Perhaps it would make sense to set a boolean somewhere, maybe in the `tuple` returned by `get_quota_status`, named something like `no_quota` or `unlimited_jobs`, to be the single source of truth for whether the user has a quota, since right now there are three variables (`max_jobs, previous_jobs, remaining_jobs`) that are all `None` when the user has no quota, and all `not None` otherwise, so it's not obvious which one of those three should be checked to determine if the user has a quota.

asjohnston-asf (approved):

PR #1016, opened 2022-05-18
Fix get jobs waiting for execution

jtherrmann:


asjohnston-asf:
FYI, this gives us similar behavior to way back in v1.1.6 before we introduced the limit 400-record limit: https://github.com/ASFHyP3/hyp3/blob/v1.1.6/apps/start-execution/src/start_execution.py#L21

It re-introduces the risk that when there are many PENDING jobs we'll spend a long time paging through them before finding 400 results, so we could see lambda timeouts. I think the consequences aren't as severe as they used to be; in the old days it prevented *any* new executions from being started, where now new executions will eventually be started as others transition from PENDING to RUNNING.

TLDR: looks good, but in the future we should consider a secondary index on `execution_started` to make this query more effective

asjohnston-asf (approved):

PR #1018, opened 2022-05-18
change security environment for hyp3-a19-jpl from JPL to JPL-public

asjohnston-asf:
The only impact to the rendered template should be to include the public bucket policy at https://github.com/ASFHyP3/hyp3/blob/develop/apps/main-cf.yml.j2#L307

Note that this leaves the `JPL` security environment option unused by any deployments at the moment. Do we anticipate needing it in the future? Should we consider consolidating `JPL` and `JPL-public`?

@jhkennedy have you validated our ability to create a public bucket policy in the a19-jpl account?

### DIFF
--- a/.github/workflows/deploy-enterprise.yml
+++ b/.github/workflows/deploy-enterprise.yml
@@ -68,7 +68,7 @@ jobs:
             default_max_vcpus: 1600
             expanded_max_vcpus: 1600
             required_surplus: 0
-            security_environment: JPL
+            security_environment: JPL-public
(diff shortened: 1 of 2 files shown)


jhkennedy:
here's the cloud ticket update showing the've remove the block access policy:
image

jhkennedy:
> Do we anticipate needing it in the future? Should we consider consolidating JPL and JPL-public?

I do now know of any upcoming deployments to a new JPL account, which `JPL` would be needed for, but I wouldn't be surprised to get a new deployment ask soon. I'm on the fence whether to rip it out now and put it back when we need it, or just leave it.

jhkennedy (approved):

PR #1025, opened 2022-05-19
Update README.md

asjohnston-asf:

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,15 +9,14 @@
 A processing environment for HyP3 Plugins in AWS.
 
 ## Table of contents
-- [Local Development](#local-development)
+- [Developer Setup](#developer-setup)
 - [Deployment](#deployment)
   - [Prerequisites](#prerequisites)
   - [Stack Parameters](#stack-parameters)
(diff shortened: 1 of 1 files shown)


jhkennedy (approved):

README.md
```diff
@@ -9,13 +9,30 @@
...
+1. Clone the repository
+   ```
+   git clone git@github.com:ASFHyP3/hyp3.git
+   cd hyp3
+   ```
+2. Create and activate a conda environment
+   ```
+   conda env create -f environment.yml
+   conda activate hyp3
+   ```
```
jtherrmann, line 30: Is `make install` no longer necessary?

asjohnston-asf: Are you asking if we should remove the `install` command from the Makefile?

jtherrmann: @asjohnston-asf No, but that's an interesting idea. Or maybe the `install` target should just run `conda env create -f environment.yml`. But that's probably beyond the scope of this PR.

jtherrmann (approved):

PR #1051, opened 2022-06-07
Use default service-linked role for AWS Batch Compute Environment

asjohnston-asf:
When HyP3 was initially developed, we had to provision an IAM role for the `ServiceRole` attribute of the AWS Batch Compute environment.

In Apr 2021, AWS Batch [introduced support](https://aws.amazon.com/about-aws/whats-new/2021/04/aws-batch-simplifies-permissions-introducing-service-linked-roles/) for Service-Linked roles. If no customer-managed `ServiceRole` is specified for a compute environment, the service-linked role is automatically created (if necessary) and assigned, which automatically grants the necessary permissions by default.

https://docs.aws.amazon.com/batch/latest/userguide/using-service-linked-roles.html
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-batch-computeenvironment.html#cfn-batch-computeenvironment-servicerole

Removing the `ServiceRole` attribute *isn't enough* to update the Service Role on existing Compute Environments. I've updated the logical resource ID for the compute environment to force a new one to be created (with the default service-linked role attached) as part of this deployment.

This ultimately addresses a "Following fields ... subnets ... can be updated only for Non-fargate Compute Environment having a Batch Service Linked Role" error when attempting to add additional subnets for an existing HyP3 deployment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [2.18.1]
 ### Changed
 - Increased `MemorySize` for `process-new-granules` function to improve performance when evaluating subscriptions.
-- Decoupled Batch Compute Environment from customer-managed service role in favor of AWS-managed service-linked role.
+- Decoupled Batch Compute Environment from customer-managed service role in favor of the default AWS-managed
+  service-linked role.
 
 ## [2.18.0]
 ### Removed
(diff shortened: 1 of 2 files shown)


asjohnston-asf:
@jhkennedy we should confirm there won't be any permissions issues generating the service-linked role in our JPL-managed AWS accounts before signing off on this PR.

jhkennedy:
We should be g2g; I was able to create a `AWSServiceRoleForBatch` in a JPL account:
image

jhkennedy (approved):

PR #1065, opened 2022-06-15
Allow updating subscription start datetime and AOI

jtherrmann:

### DIFF
--- a/apps/api/src/hyp3_api/api-spec/openapi-spec.yml
+++ b/apps/api/src/hyp3_api/api-spec/openapi-spec.yml
@@ -468,6 +468,7 @@ components:
       description: Area-of-interest as a WKT string.
       type: string
       format: wkt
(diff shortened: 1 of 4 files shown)


apps/api/src/hyp3_api/api-spec/openapi-spec.yml
```diff
@@ -243,8 +243,13 @@ components:
       type: object
       additionalProperties: false
       properties:
+        start:
+          $ref: "#/components/schemas/datetime"
         end:
           $ref: "#/components/schemas/datetime"
+        intersectsWith:
+          type: string
+          format: wkt
```
asjohnston-asf, line 252: leaning slightly towards creating a stand-alone component for `intersectsWith` to avoid a tiny bit of duplication between updating/creating a subscription

asjohnston-asf (approved):

PR #1093, opened 2022-06-29
Add enterprise test deployment

jtherrmann:

### DIFF
--- a/.github/workflows/deploy-enterprise.yml
+++ b/.github/workflows/deploy-enterprise.yml
@@ -72,6 +72,26 @@ jobs:
             ami_id: /aws/service/ecs/optimized-ami/amazon-linux-2/recommended/image_id
             distribution_url: ''
 
+          - environment: hyp3-test-enterprise
+            domain: TODO
+            template_bucket: cf-templates-aubvn3i9olmk-us-west-2
+            image_tag: test
+            product_lifetime_in_days: 14
+            quota: 0
+            job_files: >-
+              job_spec/AUTORIFT.yml
+              job_spec/INSAR_GAMMA.yml
+              job_spec/RTC_GAMMA.yml
+              job_spec/INSAR_ISCE.yml
+              job_spec/WATER_MAP.yml
+            instance_types: r5d.xlarge,r5dn.xlarge
+            default_max_vcpus: 640
+            expanded_max_vcpus: 640
+            required_surplus: 0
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-enterprise.yml
```diff
@@ -72,6 +72,27 @@ jobs:
             ami_id: /aws/service/ecs/optimized-ami/amazon-linux-2/recommended/image_id
             distribution_url: ''
 
+          - environment: hyp3-test-enterprise
+            domain: hyp3-test-enterprise.asf.alaska.edu
+            template_bucket: cf-templates-aubvn3i9olmk-us-west-2
+            image_tag: test
+            product_lifetime_in_days: 14
+            quota: 0
+            deploy_ref: refs/heads/develop
```
jhkennedy, line 81: Since this workflow is only triggered
```yaml
on:
  push:
    branches:
      - main
```
you'll need to:
- [ ]  add `develop` to the triggers
- [ ] add a `deploy_ref` to every matrix element / environment
- [ ] add the deploy ref condition:
       https://github.com/ASFHyP3/hyp3/blob/develop/.github/workflows/deploy-daac.yml#L74

jhkennedy (commented):
The values all look great!

.github/workflows/deploy-enterprise.yml
jhkennedy: Actually, thinking more about this, I think we might want to put this in a separate `deploy-enterprise-test.yml` workflow -- the way we have it setup with the deploy ref, it'll "try" to deploy to every environment but quickly succeed by skipping the deploy job if the ref != to the one specified.

That'd mean, we'd be asked for approval to deploy the prod enterprise stacks on any merge to main _or_ develop, which is rather undesirable. 

Putting it in a separate workflow file only triggered on `develop` (and not requiring reviews in the action env) I think is the better option.

jhkennedy: Another thought, we could keep it in the same workflow and see how this env setting plays out:
image

I *think* that'd just make the deploy jobs fail if not on the deployment branch, but it's possible it skips instead...

jtherrmann: @jhkennedy How do I implement:
> not requiring reviews in the action env

jhkennedy (approved):

PR #1097, opened 2022-06-30
Temporarily pin AVO hyp3-gamma image tag

jhkennedy:
AVO processing identified some issues with InSAR GAMMA products where an un-physical subsidence was observed due to GAMMA keying in on a "signal" over the water. Our current water mask is buffered such that these narrow water bodies aren't masked out.

This pins the hyp3-gamma container image to the `5.4.2.dev2_ga6225f8` tags, which was produced via ASFHyP3/hyp3-gamma#379 and uses a stricter water mask that is not buffered away from the coast line.

IMPORTANTLY, this should be a temporary change until we integrate a better water-masking workflow that's suitable to general HyP3 (see ASFHyP3/hyp3-gamma#379 for more details).

### DIFF
--- a/.github/workflows/deploy-enterprise.yml
+++ b/.github/workflows/deploy-enterprise.yml
@@ -75,7 +75,7 @@ jobs:
           - environment: hyp3-avo
             domain: hyp3-avo.asf.alaska.edu
             template_bucket: cf-templates-1x4a21iq1cba7-us-west-2
-            image_tag: latest
+            image_tag: 5.4.2.dev2_ga6225f8
             product_lifetime_in_days: 365000
             quota: 0
             job_files: job_spec/INSAR_GAMMA.yml


jhkennedy:
I was able to produce in InSAR product from the scene pair AVO identified that is strictly water masked using this image:

```sh
docker run --rm \
   -v ${HOME}/.aws:/home/conda/.aws \
   -e AWS_PROFILE=sandbox \
   ${HYP3_REGISTRY}/hyp3-gamma:5.4.2.dev2_ga6225f8 \
   ++process insar \
   --username "${EDL_USERNAME}" --password "${EDL_PASSWORD}" \
   --bucket jhk-shenanigans --bucket-prefix yitian-chen \
   --include-dem true  \
   --include-displacement-maps true \
   --include-look-vectors true \
   --include-wrapped-phase true \
   --include-inc-map true \
   --apply-water-mask true \
   --looks 20x4 \
   S1A_IW_SLC__1SDV_20200711T023803_20200711T023830_033401_03DEBA_68E0 \
   S1A_IW_SLC__1SDV_20210706T023809_20210706T023836_038651_048F95_F3E8
```
The output product is here in the Tools Sandbox account:
```
s3://jhk-shenanigans/yitian-chen/S1AA_20200711T023803_20210706T023809_VVP360_INT80_G_weF_6A1E.zip
```

Which does not show the subsidence, so I think we're good to go here! Marking ready for review

jtherrmann (approved):

PR #1098, opened 2022-07-01
scale based on total spending rather than ec2 spending

asjohnston-asf:
Will need to re-create the MONTHLY_COMPUTE_BUDGET repository secret as MONTHLY_BUDGET. Might be a nice time to turn that into an environment secret; I think it's only used by `hyp3-edc-uat` and `hyp3-edc-prod` (the only environments where default_max_vcpus != expanded_max_vcpus).


asjohnston-asf:
I've created MONTHLY_BUDGET secrets for the hyp3-edc-uat and hyp3-edc-prod environments. After this feature is released and deployed to production we can delete the MONTHLY_COMPUTE_BUDGET repo secret.

jtherrmann (approved):

PR #1101, opened 2022-07-01
temporarily disable compute environment when needed during a scale-down event

asjohnston-asf:


jhkennedy (approved):
Oh, nice, this is a good way to handle it. What the reasoning behind 30 minutes? (not that I disagree with it)

asjohnston-asf:
An execution might leave the compute environment in a DISABLED state; I figured 30 minutes would be a good length of time to wait before checking back in.

asjohnston-asf:
In this situation in practice, we'll typically have a full fleet of 2,000 vCpus and be trying to scale down to 1,000. We'd have to wait for half the in-progress jobs to finish. Average job time is ~60 minutes, so the sweet spot is likely closer to 35 minutes.

PR #1117, opened 2022-07-18
Upped timeout to 5.5 hours

jacquelynsmale:


jacquelynsmale:
@asjohnston-asf it was based on 30m on my Mac, which runs a bit faster than an EC2 deployment.

asjohnston-asf (approved):

PR #1181, opened 2022-09-07
Add in Landsat 4, 5, 7 as autorift option

jacquelynsmale:


job_spec/AUTORIFT.yml
```diff
@@ -31,12 +31,18 @@ AUTORIFT:
               minLength: 23
               maxLength: 25
               example: S2A_22WEB_20200627_0_L1C
-            - description: The name of the Landsat 8 or 9 Collection 2 granule to process
+            - description: The name of the Landsat 7, 8 or 9 Collection 2 granule to process
               type: string
-              pattern: "^L[CO]0[89]_"
+              pattern: "^L[ECO]0[789]_"
```
asjohnston-asf, line 36: As written this allows LC07, LO07, LE08, LE09 which are combinations which don't exist, so I doubt anyone will use them, but I think it's worth being more specific so we remember what our intent was when we look back a few years down the road. We also could add entries to `test_submit_bad_rtc_granule_names` to make it extra clear those aren't valid options.

I think all five platforms can be handled with one pattern of:
```
pattern: "^L([CO]0[89]|E07|[TM]0[45])_"
```
- C/O for 08 and 09
- E for 07
- T/M for 04 and 05

But that's a little dense, and the fact that I felt the need to write out the details in prose suggests trying something more readable. We could do separate entries with separate patterns for each of 89, 7, 45:
```
pattern: "^L[CO]0[89]_"
pattern: "^LE07_"
pattern: "^L[TM]0[45]_"
```
Thoughts? I don't know that this mandates action, and everything else in the PR looks good.

tests/test_api/test_submit_job.py
```diff
@@ -263,6 +263,9 @@ def test_submit_good_autorift_granule_names(client, tables):
         'S2A_1UCR_20210124_0_L1C',
         'S2B_22WEB_20200612_0_L1C',
         'S2A_22XEQ_20190610_10_L1C',
+        'LM04_L1GS_025009_19830519_20200903_02_T2',
+        'LT05_L1TP_091090_20060929_20200831_02_T1',
+        'LE07_L2SP_233095_20200102_20200822_02_T2',
```
asjohnston-asf, line 268: This might be more an issue for hyp3-autorift than hyp3, but this looks like a Level 2 scene, where everything else we've dealt with is Level 1. @jhkennedy should we restrict Landsat scene names to Level 1 scenes by requiring `L1` in the 6-7th characters?

hyp3-autorift explicitly uses the L1 URL when lookup up scenes in the Landsat STAC catalog: https://github.com/ASFHyP3/hyp3-autorift/blob/develop/hyp3_autorift/process.py#L37

This is a valid L2 scene and exists in S3 at `s3://usgs-landsat/collection02/level-2/standard/etm/2020/233/095/LE07_L2SP_233095_20200102_20200822_02_T2/`, but looking it up at https://landsatlook.usgs.gov/stac-server/collections/landsat-c2l1/items/LE07_L2SP_233095_20200102_20200822_02_T2 gives "not found".

The corresponding L1 scene is at `s3://usgs-landsat/collection02/level-1/standard/etm/2020/233/095/LE07_L1GT_233095_20200102_20200822_02_T2/`, and querying https://landsatlook.usgs.gov/stac-server/collections/landsat-c2l1/items/LE07_L1GT_233095_20200102_20200822_02_T2 finds it as expected.

jhkennedy: Yes, we'll typically be processing L1 scenes as the L2 adds in some corrections that mess with the feature tracking, so I think it'd be worth restricting to Level 1 scenes. 


For reference, here are the L7 test pairs we've been using to develop the wallis filter:
```
LE07_L1TP_063018_20040911_20200915_02_T1,LE07_L1TP_063018_20040810_20200915_02_T1
LE07_L1TP_061018_20050324_20200914_02_T1,LE07_L1TP_061018_20040321_20200915_02_T1
LE07_L1TP_061018_20040625_20200915_02_T1,LE07_L1TP_061018_20030810_20200916_02_T1
LE07_L1TP_061018_20040711_20200915_02_T1,LE07_L1TP_061018_20040625_20200915_02_T1
LE07_L1TP_061018_20050511_20200914_02_T1,LE07_L1TP_061018_20040625_20200915_02_T1
LE07_L1TP_061018_20040711_20200915_02_T1,LE07_L1TP_061018_20030810_20200916_02_T1
LE07_L1TP_061018_20040625_20200915_02_T1,LE07_L1TP_061018_20040321_20200915_02_T1
LE07_L1TP_061018_20050425_20200914_02_T1,LE07_L1TP_061018_20040321_20200915_02_T1
LE07_L1TP_062018_20050416_20200914_02_T1,LE07_L1TP_062018_20040803_20200915_02_T1
LE07_L1TP_063018_20050509_20200914_02_T1,LE07_L1TP_063018_20040810_20200915_02_T1
LE07_L1TP_062018_20040803_20200915_02_T1,LE07_L1TP_062018_20040718_20200915_02_T1
```

job_spec/AUTORIFT_ITS_LIVE.yml
```diff
@@ -31,18 +31,12 @@ AUTORIFT:
               minLength: 23
               maxLength: 25
               example: S2A_22WEB_20200627_0_L1C
-            - description: The name of the Landsat 7, 8 or 9 Collection 2 granule to process
+            - description: The name of the Landsat 4, 5, 7, 8 or 9 Collection 2 granule to process
               type: string
-              pattern: "^L[ECO]0[789]_"
+              pattern: "^L([CO]0[89]|E07|T0[45])_.*_T2"
```
jhkennedy: I don't think we want the `_T2` here

tests/test_api/test_submit_job.py
jacquelynsmale: I looked through the ITS_LIVE Landsat scene name bucket (s3://its-live-data/NSIDC/v01/velocity_image_pair) and confirmed that there are just L1 data there. I added a pattern to the check in job-spec/AUTRIFT*.yml and a failure case in `test_submit_bad_autorift_granule_names` (L305 in tests/test_api/test_submit_job.py).

job_spec/AUTORIFT_ITS_LIVE.yml
jacquelynsmale: Ope yea I realized that after submitting it, so now there's an updated string as `"^L([CO]0[89]|E07|T0[45])_L1"`

asjohnston-asf (approved):

jhkennedy (approved):

PR #1189, opened 2022-09-09
upgrade moto to v4.0.2

asjohnston-asf:
I haven't quite figured out *why* this particular configuration works, but it works. So far it only works if I include both the `@mock_dynamodb` decorator *and* the `with mock_dynamodb()` context manager.

See "pesky imports" at http://docs.getmoto.org/en/latest/docs/getting_started.html#what-about-those-pesky-imports

### DIFF
--- a/requirements-all.txt
+++ b/requirements-all.txt
@@ -6,7 +6,7 @@
 -r requirements-apps-update-db.txt
 boto3==1.24.68
 jinja2==3.1.2
-moto==1.3.14
+moto[dynamodb]==4.0.2
 pytest==7.1.3
 PyYAML==6.0
 responses==0.21.0
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,7 +9,5 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
-    ignore:
-      - dependency-name: "moto"
     labels:
       - "bumpless"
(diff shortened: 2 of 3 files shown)


asjohnston-asf:
@jtherrmann @jhkennedy I think this is a step forward, even if I'm not clear on the details? What are your thoughts on merging as-is?

requirements-all.txt
```diff
@@ -6,7 +6,7 @@
 -r requirements-apps-update-db.txt
 boto3==1.24.70
 jinja2==3.1.2
-moto==1.3.14
+moto[dynamodb]==4.0.2
```
jtherrmann, line 9: Just out of curiosity what does the `[dynamodb]` do?

asjohnston-asf: It filters which AWS services are installed, so limiting to just the `dynamodb` service we use to reduce the overall install footprint

http://docs.getmoto.org/en/latest/docs/getting_started.html
https://stackoverflow.com/questions/46775346/what-do-square-brackets-mean-in-pip-install

jtherrmann: Nevermind, found it at http://docs.getmoto.org/en/latest/docs/getting_started.html

jtherrmann:
https://github.com/ASFHyP3/hyp3/issues/1192

jtherrmann (approved):
I don't think there's a clear way to improve on this at the moment. See the issue linked above.

PR #1203, opened 2022-09-16
upgrade to latest generation r6id/c6id instance types

asjohnston-asf:
These instance types were made available earlier this year and AWS Batch added support on on 2022-09-15.
- https://aws.amazon.com/about-aws/whats-new/2022/05/introducing-amazon-ec2-c6id-instances/
- https://aws.amazon.com/about-aws/whats-new/2022/06/introducing-amazon-r6id-instances/

A new generation of the network-optimized `r5dn` instance type is not (yet) available.

Overall this should reduce both the processing time and processing cost for HyP3 jobs, possibly by 10-20%. I've attached the spot price history for the relevant instance types for comparison. The [AWS Spot Advisor]() shows the new instance types have seen similar interruption rates over the last month.

Using the new instance types exclusively should deliver the lowest costs, but they may introduce more volatility. We could instead *add* the new instance types to our existing pool, only realizing a portion of the savings but greatly reducing volatility. Adding an option to set the Spot [BidPercentage](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-computeenvironment-computeresources.html#cfn-batch-computeenvironment-computeresources-bidpercentage) attribute could give us the best of both worlds.

## Compute Optimized
Screenshot from 2022-09-16 08-31-11
Screenshot from 2022-09-16 08-18-04
Screenshot from 2022-09-16 08-18-18

Screenshot from 2022-09-16 08-19-30
Screenshot from 2022-09-16 08-19-40

## Memory Optimized (xlarge)
Screenshot from 2022-09-16 08-31-11
Screenshot from 2022-09-16 08-20-58
Screenshot from 2022-09-16 08-21-05
Screenshot from 2022-09-16 08-21-18

Screenshot from 2022-09-16 08-21-46
Screenshot from 2022-09-16 08-21-56
Screenshot from 2022-09-16 08-22-07

## Memory Optimized (4xlarge)
Screenshot from 2022-09-16 08-31-11
Screenshot from 2022-09-16 08-22-52
Screenshot from 2022-09-16 08-23-00
Screenshot from 2022-09-16 08-23-13

Screenshot from 2022-09-16 08-23-41
Screenshot from 2022-09-16 08-23-52
Screenshot from 2022-09-16 08-24-02

### DIFF
--- a/.github/workflows/deploy-daac.yml
+++ b/.github/workflows/deploy-daac.yml
@@ -22,7 +22,7 @@ jobs:
             quota: 1000
             deploy_ref: refs/heads/main
             job_files: job_spec/AUTORIFT.yml job_spec/INSAR_GAMMA.yml job_spec/RTC_GAMMA.yml
-            instance_types: r6id.xlarge
+            instance_types: r6id.xlarge,r5dn.xlarge
             default_max_vcpus: 1000
             expanded_max_vcpus: 2000
             required_surplus: 2000
@@ -44,7 +44,7 @@ jobs:
               job_spec/RTC_GAMMA.yml
               job_spec/INSAR_ISCE.yml
               job_spec/WATER_MAP.yml
-            instance_types: r6id.xlarge
+            instance_types: r6id.xlarge,r5dn.xlarge
             default_max_vcpus: 1000
             expanded_max_vcpus: 2000
             required_surplus: 2000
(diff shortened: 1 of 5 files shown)


jhkennedy:
I think cutting over is probably fine for most of our deployments, but might be worth adding to the fleet instead for our production deployments. 

BidPercentage sounds like a good idea overall, but we don't necessarily have to do that here.

asjohnston-asf:
`r6id.xlarge` is the only instance type where I'm concerned about potential volatility. I've kept `r5dn.xlarge` as an option in all of those deployments to hedge against that risk.

jtherrmann (approved):

PR #1215, opened 2022-09-22
Parallelize subscription handling using manager/worker model

jtherrmann:
TODO:
- [x] Do we like the basic approach?
- [x] Is it fine if the worker times out while processing subscriptions?
- [x] Address these issues (failed events, duplicate events, permissions, etc.):
  - https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#Lambda.Client.invoke
    - See "For asynchronous invocation , Lambda [....]"
    - ~~"This operation requires permission for the lambda:InvokeFunction action"~~
- [x] Finish tests
- [x] Add/update cloudformation stuff
- [x] Rename the two lambda functions to something more clear, e.g. `subscriptions_manager` and `subscriptions_worker`, or `invoke_process_subscriptions` and `process_subscriptions`, etc. (I'd like to include `subscriptions` in the names as I've always felt that's missing from `process_new_granules`.)
- [x] See remaining `TODO` items in code.


apps/process-new-granules/src/process_new_granules.py
```diff
@@ -1,83 +1,31 @@
...
-    processed_jobs, _ = dynamo.jobs.query_jobs(
-        user=subscription['user_id'],
-        name=subscription['job_specification']['name'],
-        job_type=subscription['job_specification']['job_type'],
+def handle_subscription_batch(subscriptions: list[dict], cutoff_date: datetime, worker_function_arn: str) -> None:
+    payload = {
+        'subscriptions': subscriptions,
+        'cutoff_date': cutoff_date.isoformat(),
+    }
+    client = boto3.client('lambda')
```
asjohnston-asf, line 14: Err on the side of declaring boto3 clients in global space, so many api calls (and subsequent lambda executions) can share the same http session:

https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html
> Take advantage of execution environment reuse to improve the performance of your function. Initialize SDK clients and database connections outside of the function handler, and cache static assets locally in the /tmp directory. Subsequent invocations processed by the same instance of your function can reuse these resources. This saves cost by reducing function run time.
> 
> To avoid potential data leaks across invocations, don’t use the execution environment to store user data, events, or other information with security implications. If your function relies on a mutable state that can’t be stored in memory within the handler, consider creating a separate function or separate versions of a function for each user.

apps/process-new-granules/src/process_new_granules.py
```diff
@@ -1,83 +1,31 @@
...
-        cutoff_date = datetime.now(tz=timezone.utc) - timedelta(days=5)
-        if subscription['enabled']:
-            handle_subscription(subscription)
-
-            if dateutil.parser.parse(subscription['search_parameters']['end']) <= cutoff_date\
-                    and len(get_unprocessed_granules(subscription)) == 0:
-                disable_subscription(subscription)
+    cutoff_date = datetime.now(tz=timezone.utc) - timedelta(days=5)
+    enabled_subscriptions = [subscription for subscription in subscriptions if subscription['enabled']]
+    batch_size = 3
```
asjohnston-asf, line 29: I don't see a need to batch the worker calls, I think it's a simpler and more efficient interface to have the worker take a single subscription id

apps/process-new-granules/src/process_new_granules.py
```diff
@@ -1,83 +1,31 @@
...
     subscriptions = dynamo.subscriptions.get_all_subscriptions()
-    for subscription in subscriptions:
-        cutoff_date = datetime.now(tz=timezone.utc) - timedelta(days=5)
-        if subscription['enabled']:
-            handle_subscription(subscription)
-
-            if dateutil.parser.parse(subscription['search_parameters']['end']) <= cutoff_date\
-                    and len(get_unprocessed_granules(subscription)) == 0:
-                disable_subscription(subscription)
+    cutoff_date = datetime.now(tz=timezone.utc) - timedelta(days=5)
```
asjohnston-asf, line 27: I think the details of the cutoff date are specific to the worker, and could be encapsulated in the other lambda function

apps/process-new-granules/src/process_new_granules.py
```diff
@@ -1,83 +1,31 @@
...
-    dynamo.subscriptions.put_subscription(subscription['user_id'], subscription)
-
-
-def handle_subscription(subscription):
-    print(f'Handling subscription {subscription["subscription_id"]} for user {subscription["user_id"]}')
-    jobs = get_jobs_for_subscription(subscription, limit=20)
-    if jobs:
-        print(f'Submitting {len(jobs)} jobs')
-        dynamo.jobs.put_jobs(subscription['user_id'], jobs, fail_when_over_quota=False)
+    # TODO attempt to log worker success or failure?
```
asjohnston-asf, line 20: `InvocationType='Event'` is a "fire and forget" model. `response` will say "I was able to start the execution", but not "the execution succeeded/failed. I think it's sufficient to log "started worker execution for {subscription_id}" here.

apps/process-new-granules/src/process_new_granules.py
jtherrmann: I agree it's simpler, but I'm interested to hear why you think it would be more efficient.

jtherrmann:
@asjohnston-asf I think you'd mentioned passing the worker a subscription ID, but it's slightly simpler to just pass the worker a subscription object, since the manager already fetches all subscriptions and filters out the disabled ones. See the [subscription manager lambda handler](https://github.com/ASFHyP3/hyp3/blob/f72ecd0d3602202a5a32473b21f83074e80a8335/apps/subscription-manager/src/subscription_manager.py#L21). Do you have any objection to this? Is there a particular reason you'd prefer to pass just the subscription ID to the worker and have it re-fetch the subscription? (Only reason I can think of would be if the subscription gets updated while the worker is being launched, but that seems unlikely.)

asjohnston-asf:
That sounds solid. Makes the worker easier to test, too, if it takes a whole dictionary as input instead of relying on a database lookup. I hadn't realized the manager already has all the subscription-specific information on hand.

jtherrmann:
@asjohnston-asf With regard to [async lambda invocations](https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html), I think this summarizes the results of our previous discussion:
- We're going to assume that the default concurrency limit is sufficient for our current number of subscriptions, but we may have to revisit this assumption in the future. See the paragraph "If the function doesn't have enough concurrency available to process all events [....]" from the above source. Also see [Lambda function scaling](https://docs.aws.amazon.com/lambda/latest/dg/invocation-scaling.html).
- We're not going to worry about lost events or failed invocations, as those subscriptions should just get re-processed the next time the manager runs.
  - We're going to set the worker's max retry attempts to `0`, for the same reason.
- We're not going to worry about duplicate events, because the worst-case scenario is a few duplicate hyp3 jobs for a subscription.
  - If this does become a problem, it shouldn't be too hard to come up with a way to filter out duplicate events.

Does that all sound correct?

tests/test_subscription_worker.py
```diff
@@ -3,8 +3,7 @@
 
 import asf_search
 import pytest
-
-import process_new_granules
+import subscription_worker
```
asjohnston-asf, line 6: `subscription_worker` should be an application import in it's own paragraph?

Makefile
```diff
@@ -39,7 +42,7 @@ render:
 static: flake8 openapi-validate cfn-lint
 
 flake8:
-	flake8 --ignore=E731 --max-line-length=120 --import-order-style=pycharm --statistics --application-import-names hyp3_api,get_files,handle_batch_event,check_processing_time,start_execution,update_db,upload_log,dynamo,process_new_granules,scale_cluster apps tests lib
+	flake8 --ignore=E731 --max-line-length=120 --import-order-style=pycharm --statistics --application-import-names hyp3_api,get_files,handle_batch_event,check_processing_time,start_execution,update_db,upload_log,dynamo,subscription_manager,scale_cluster apps tests lib
```
asjohnston-asf, line 45: Add `subscription_worker` here?

jtherrmann:
If manager and worker both have timeout (max runtime) of 15 min, and the manager
 runs every 35 min, and invocation events time out after 4 min, then there shoul
d be no overlap between two workers processing the same subscription, e.g:

- manager starts at 0m
- manager invokes `worker-x` to process `sub-A` at 15m
- invocation event waits in queue for 4m
- `worker-x` starts at 19m
- `worker-x` runs until 34m
- manager starts again at 35m
- manager invokes `worker-y` to process `sub-A` at 35m

So `worker-x` and `worker-y` never overlap.

asjohnston-asf (dismissed):

asjohnston-asf (approved):

PR #1238, opened 2022-10-06
automate updates of github action versions

asjohnston-asf:
copied from the already-working dependabot.yml at https://github.com/ASFHyP3/actions/blob/develop/.github/dependabot.yml

We now also pin all github actions to specific versions in their `uses: ` fields. Currently pins are a mix of major and patch versions. Should we consider everything to a patch version, or maybe only to a minor version?


jhkennedy (approved):
The reusable ones need to be fully specified, but I think it's fine to leave the marketplace ones broad for now

PR #1247, opened 2022-10-11
add temporary hyp3-edc-contingency environment to deploy-daac.yml

asjohnston-asf:
Should we actually merge this into `develop`, or should we use an alternate `deploy_ref` so we don't have a merge/revert in our git history?

Corresponding hyp3-edc-contingency github environment has been created.


jhkennedy (approved):

jhkennedy:
Develop and revert out is fine by me -- no need to be vain with he git history.

PR #1258, opened 2022-10-12
add new execution_started index to jobs table

asjohnston-asf:
Should we include a `RANGE` key on `request_time` to the new index as well? It's not strictly necessary; the `status_code` index we're replacing didn't have one, but adding one would let us prioritize starting executions in the order that they were submitted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
-## [2.21.3]
+## [2.21.4]
 ### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [2.21.3]
```
jtherrmann, line 8: This is identical to the previous version.

jtherrmann (approved):

jtherrmann:
Actually yes, I think I'd like to see that `RANGE` key added as well.

jtherrmann (approved):

PR #1261, opened 2022-10-13
Revert "re-add execution_started index with correct attribute type"

asjohnston-asf:
Reverts ASFHyP3/hyp3#1260

This didn't work either. These queries always return zero results:
```
response = table.query(IndexName='execution_started', KeyConditionExpression=Key('execution_started').eq('false'))
response = table.query(IndexName='execution_started', KeyConditionExpression=Key('execution_started').eq('true'))
```
And querying using a boolean type throws an error because the index is expecting a string:
```
response = table.query(IndexName='execution_started', KeyConditionExpression=Key('execution_started').eq(False))

botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the Query operation: One or more parameter values were invalid: Condition parameter type does not match schema type
```

Google suggests it's simply not possible to create in index on a boolean attribute. https://stackoverflow.com/questions/28258024/dynamodb-query-on-boolean-key has some discussion of alternatives.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-
-## [2.21.4]
-### Added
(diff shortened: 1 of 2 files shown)


jtherrmann (approved):
:cry:

PR #1263, opened 2022-10-13
Parallelize StartExecution using manager/worker model

jtherrmann:
- [x] Update files (e.g. CloudFormation) and tests
- [x] See `TODO`s in code
- [x] Disable the old StartExecution schedule before deployment (see Andrew's comment below)


apps/start-execution-manager/src/start_execution_manager.py
```diff
@@ -0,0 +1,30 @@
...
+        FunctionName=worker_function_arn,
+        InvocationType='Event',
+        Payload=payload,
+    )
+
+
+# TODO add logging
+def lambda_handler(event, context):
+    worker_function_arn = os.environ['SUBSCRIPTION_WORKER_ARN']  # TODO add this in cf
+    pending_jobs = dynamo.jobs.get_jobs_waiting_for_execution(limit=400)  # TODO increase this limit, or just remove it?
```
jtherrmann, line 27: limit 900 and submit batches of 300?

jtherrmann:
Referencing https://github.com/ASFHyP3/hyp3/pull/1215#issuecomment-1263885173, if the manager has timeout 10 seconds and runs every 2 minutes, and the worker has timeout 45 seconds and max event age 60 seconds, then there should be no overlap between two workers starting executions for the same job(s), e.g:

- manager starts at 0s
- manager invokes a worker at 10s
  - invocation event waits in queue for 60s
- worker starts at 70s
- worker runs until 115s
- manager starts again at 120s

It wouldn't actually be a problem if two workers attempted to start executions for the same job, because the second attempt would fail as described [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/stepfunctions.html#SFN.Client.start_execution), but it would be an inefficient use of the worker's resources.

asjohnston-asf:
FYI, when deploying these changes, the new StartExecutionManager lambda and schedule will be created before the old StartExecution lambda and schedule are deleted, so there may be a window during deployment where they're both running. We might consider disabling the old StartExecution schedule before deployment, but I don't *think* having both run at the same time will cause any harm.

tests/test_subscription_manager.py
```diff
@@ -17,7 +17,7 @@ def test_lambda_handler(tables):
         tables.subscriptions_table.put_item(Item=item)
 
     with patch('subscription_manager.invoke_worker') as mock_invoke_worker, \
-            patch.dict(os.environ, {'SUBSCRIPTION_WORKER_ARN': TEST_WORKER_ARN}):
+            patch.dict(os.environ, {'SUBSCRIPTION_WORKER_ARN': TEST_WORKER_ARN}, clear=True):
```
asjohnston-asf, line 20: `clear=True` breaks the test, since it discards the required `SUBSCRIPTIONS_TABLE_NAME` variable that's currently set in `cfg.env`. I think I like setting env vars for testing in the tests themselves, but that risks introducing these kinds of conflicts now that we're setting env vars in multiple places.

Dropping `clear=True` lets the test pass for now, if we want to do that to move forward, but should we use this opportunity to move some values out of cfg.env?

apps/start-execution-manager/src/start_execution_manager.py
```diff
@@ -0,0 +1,39 @@
...
+
+    pending_jobs = dynamo.jobs.get_jobs_waiting_for_execution(limit=1800)
+    logger.info(f'Got {len(pending_jobs)} pending jobs')
+
+    batch_size = 300
+    for i in range(0, len(pending_jobs), batch_size):
+        jobs = pending_jobs[i:i + batch_size]
+        logger.info(f'Invoking worker for {len(jobs)} jobs')
+        response = invoke_worker(worker_function_arn, jobs)
+        logger.info(f'Got response status code {response["StatusCode"]}')
```
asjohnston-asf, line 39: Adding this log statement broke some tests, since it's interacting with the output of a mocked `invoke_worker` call. I'm content to drop the log statement if we want to take the path of least resistance.

jtherrmann: Ah right, I meant to mock that return value.

tests/test_subscription_manager.py
jtherrmann: Oh that's probably why I didn't include it originally, let's just drop it for now.

asjohnston-asf (approved):

PR #1271, opened 2022-10-17
reduce worker concurrency by running manager twice as often for half the jobs

asjohnston-asf:
@jtherrmann Here's a draft of the straightforward way to handle the batch.submit_job service limit; I imagine there's more discussion to be had but I at least wanted to see what this version looked like.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [2.21.4]
 ### Changed
 - Batches of step function executions are now started in parallel using a manager to launch one worker per batch of jobs
-  (currently up to 6 batches of 300 jobs for a total of 1800 jobs each time the manager runs).
+  (currently up to 3 batches of 300 jobs for a total of 900 jobs each time the manager runs).
 
 ## [2.21.3]
 ### Added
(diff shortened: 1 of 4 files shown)


jtherrmann:
Referencing https://github.com/ASFHyP3/hyp3/pull/1263#issuecomment-1279311791, if the manager has timeout 10 seconds and runs every 1 minute, and the worker has timeout 45 seconds and max event age 60 seconds, then there **may be some overlap** between two workers starting executions for the same job(s), e.g:

* manager starts at 0s
* manager invokes `worker-x` at 10s
    * invocation event waits in queue for 60s
* manager starts again at 60s and invokes `worker-y`
* `worker-x` starts at 70s
* `worker-x` and `worker-y` are now running at the same time and may attempt to start the step function for the same job(s)

Note the behavior of `StartExecution` for duplicate requests as described [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/stepfunctions.html#SFN.Client.start_execution):

> `StartExecution` is idempotent. If `StartExecution` is called with the same name and input as a running execution, the call will succeed and return the same response as the original request. If the execution is closed or if the input is different, it will return a 400 `ExecutionAlreadyExists` error. Names can be reused after 90 days.

It seems the most likely scenario for duplicate requests in this case is that the original execution is still running, in which case the second request should do nothing, which is inefficient but won't cause the entire batch of job submissions to fail. We should just make sure that we're getting the throughput we expect, and maybe query the logs for `ExecutionAlreadyExists` errors.

jtherrmann (approved):

PR #1276, opened 2022-10-18
add AUTORIFT_TEST job type to hyp3-its-live deployment

asjohnston-asf:
We want to run a non-trivial amount of data processing to validate the `hyp3-autorift:test` container before a production release. This PR adds the ability to run jobs using the test container to the hyp3-its-live.asf.alaska.edu deployment so the validation jobs can be run there and billed to the project.

It's helpful to diff `AUTORIFT_ITS_LIVE_TEST.yml` with `AUTORIFT_ITS_LIVE.yml`; the only differences should be the `AUTORIFT_TEST` job name and the `image_tag: test` field.

Does this deserve a changelog entry? I imagine we'll release it immediately, and I always lean against bumpless releases.

### DIFF
--- a/job_spec/AUTORIFT_ITS_LIVE_TEST.yml
+++ b/job_spec/AUTORIFT_ITS_LIVE_TEST.yml
@@ -1,0 +1,59 @@
+AUTORIFT_TEST:
+  required_parameters:
+    - granules
+  parameters:
+    granules:
+      default:  '""'
+      api_schema:
+        type: array
+        minItems: 2
+        maxItems: 2
+        example:
+          - S2B_13CES_20200105_0_L1C
+          - S2B_13CES_20200315_0_L1C
+        items:
+          anyOf:
+            - description: The name of the Sentinel-1 SLC granule to process
+              type: string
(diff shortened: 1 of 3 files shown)


job_spec/AUTORIFT_ITS_LIVE_TEST.yml
```diff
@@ -0,0 +1,59 @@
...
+      image_tag: test
+      command:
+        - --bucket
+        - '!Ref Bucket'
+        - --bucket-prefix
+        - Ref::bucket_prefix
+        - --parameter-file
+        - '/vsicurl/http://its-live-data.s3.amazonaws.com/autorift_parameters/v001/autorift_landice_0120m.shp'
+        - --naming-scheme
+        - ITS_LIVE_PROD
```
jhkennedy, line 55: :thinking: for test jobs, I wonder if we should keep the default ASF OD naming scheme to not confuse them with production jobs...

I guess it's a better test to use the same naming scheme here and makes golden testing easier as well.

jhkennedy (approved):

jtherrmann (approved):

Issue #1286, opened 2022-10-24
Include AWS request ID in subscription manager/worker log records for unhandled exceptions


jtherrmann:
When the subscription manager or worker lambda crashes due to an unhandled exception, the AWS request ID is not included in the corresponding log record. Run the following query in CloudWatch Logs Insights for the subscription worker log group in the `edc-prod` account (setting start date to `2022/10/01` currently yields 16 matching records) to observe the missing `@requestId` field:

```
fields @timestamp, @requestId, @message, @logStream
| filter @message like /ERROR/
```

This breaks the subscription worker error summary script in https://github.com/ASFHyP3/hyp3-operations. Perhaps there's a special AWS-provided fix for this problem (similar to how they provide built-in logging configuration for Lambda), or it should be easily resolved by just logging all unhandled exceptions (with traceback) in the lambda handlers.

Validate in test:
- [x] Unhandled exceptions are now being logged with all the usual fields, including request ID. The easiest way to validate this would be to invoke the Subscription Worker with a disabled subscription and confirm that the logs show an unhandled `ValueError`.

Validate in prod:
- [x] Same as for test.

jtherrmann:
I think I actually prefer manually closing issues following successful validation.

jtherrmann:
Unfortunately overriding `sys.excepthook` does not seem to work, see https://github.com/gkrizek/leh

jtherrmann:
After confirming this problem, if there are no better solutions for logging unhandled exceptions in AWS Lambda, one possibility is to use a decorator to wrap the lambda handler in a try/except to log unhandled exceptions.

jtherrmann:
I invoked the Subscription Worker in `hyp3-edc-uat` with:

```
aws lambda invoke --profile edc-uat --function-name hyp3-edc-uat-SubscriptionWorker-SXLTCW5M16V-Lambda-E45yTNfsjejo --payload '{"subscription": {"subscription_id": "foo", "user_id": "jth", "enabled": false}}' --cli-binary-format raw-in-base64-out /dev/stdout
```

And confirmed that the log stream looks as expected:

```
START RequestId: 80a4bc7b-a49b-418d-9d06-3ef821ee09f4 Version: $LATEST
[INFO]	2022-12-14T01:36:13.023Z	80a4bc7b-a49b-418d-9d06-3ef821ee09f4	Handling subscription foo for user jth
[...]

PR #1289, opened 2022-10-25
Log unhandled exceptions in lambdas with logging configured

jtherrmann:
See https://github.com/ASFHyP3/hyp3/issues/1286

### DIFF
--- a/apps/subscription-manager/src/subscription_manager.py
+++ b/apps/subscription-manager/src/subscription_manager.py
@@ -1,16 +1,24 @@
 import json
 import logging
 import os
+import sys
 
 import boto3
 
 import dynamo
 
+LAMBDA_CLIENT = boto3.client('lambda')
+
 logging.basicConfig()
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
 
-LAMBDA_CLIENT = boto3.client('lambda')
+
+def handle_exception(exc_type, value, traceback) -> None:
+    logger.critical('Unhandled exception', exc_info=(exc_type, value, traceback))
+
+
+sys.excepthook = handle_exception
 
 
 def invoke_worker(worker_function_arn: str, subscription: dict) -> dict:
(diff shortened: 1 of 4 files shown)


apps/subscription-manager/src/subscription_manager.py
```diff
@@ -1,16 +1,24 @@
 import json
 import logging
 import os
+import sys
 
 import boto3
 
 import dynamo
 
+LAMBDA_CLIENT = boto3.client('lambda')
```
asjohnston-asf, line 10: I also prefer to see global variables initialized before any function definitions, but should this be after the logging setup so that any exceptions it throws are logged?

asjohnston-asf: We assign a lambda to excepthook instead of a function, but I don't know that it improves things much:
`sys.excepthook = lambda exc_type, value, traceback: logger.critical('Unhandled exception', exc_info=(exc_type, value, traceback))`

asjohnston-asf (approved):

apps/subscription-manager/src/subscription_manager.py
asjohnston-asf: Approved as-is, I haven't convinced myself we can do much better, although I'm open to thoughts on the matter.

jtherrmann: Ah good point, the logging setup should probably come before anything else. Another idea I had was to move the logging boilerplate:

```python3
logging.basicConfig()
logger = logging.getLogger()
logger.setLevel(logging.INFO)


def handle_exception(exc_type, value, traceback) -> None:
    logger.critical('Unhandled exception', exc_info=(exc_type, value, traceback))


sys.excepthook = handle_exception
```

somewhere else that could be reused across all Lambda functions, so we wouldn't have to copy-paste it for every Lambda. Would it work to just move it into a separate module and then `import` that module from every Lambda? I could experiment with that solution in `tools-sandbox` if we're not sure.

jtherrmann: I went ahead and merged because I want to start logging unhandled exceptions properly, but created https://github.com/ASFHyP3/hyp3/issues/1295 to track the logging boilerplate issue.

PR #1316, opened 2022-11-08
disable processing in hyp3-tibet-jpl deployment

asjohnston-asf:
Per conversation with @cmarshak , we're suspending processing in the hyp3-tibet-jpl deployment for the time being, but aren't ready to retire the deployment and may do additional processing in future months. This PR reduces the maximum EC2 fleet size to zero so no processing can happen until we increase it again.


jhkennedy (approved):
Nice! This is a good way to do it.

PR #1331, opened 2022-11-22
Add `lambda_logging` package and log unhandled exceptions

jtherrmann:
See https://github.com/ASFHyP3/hyp3/issues/1286

The motivation behind this PR is to include the AWS request ID when logging unhandled exceptions in Lambda functions.

- [x] Test the `lambda_logging` module in a sandbox stack to see if it works as expected (still logs normal stuff correctly, logs unhandled exceptions correctly, and re-raises the exception).
- [x] Determine if we need to modify the `logging` config since it's now in a separate module rather than in the lambda handler itself.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,6 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
 lib64/
 parts/
 sdist/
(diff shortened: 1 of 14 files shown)


jtherrmann:
[This test](https://github.com/ASFHyP3/hyp3/blob/develop/tests/test_subscription_worker.py#L288) is failing because we now log unhandled exceptions and exit with status code `1`. I'd like to keep this behavior because it avoids the exception appearing twice in the logs (once with the request ID, once without). We could modify the test to assert that there's a system exit and that the expected exception appears in the log output, but probably the better way would be to disable the `log_exceptions` decorator for tests (probably with a mock).

asjohnston-asf (approved):

PR #1354, opened 2022-12-08
Increase vcpu and budget for edc prod

jtherrmann:
See https://asfdaac.atlassian.net/browse/TOOL-1070

I updated monthly budget in the environment secret for edc-prod and attached screenshots to the JIRA issue.

Edit: I also updated the secret for edc-uat.

### DIFF
--- a/.github/workflows/deploy-daac.yml
+++ b/.github/workflows/deploy-daac.yml
@@ -45,9 +45,9 @@ jobs:
               job_spec/INSAR_ISCE.yml
               job_spec/WATER_MAP.yml
             instance_types: r6id.xlarge,r5dn.xlarge,r5d.xlarge
-            default_max_vcpus: 1000
-            expanded_max_vcpus: 2000
-            required_surplus: 2000
+            default_max_vcpus: 1500
+            expanded_max_vcpus: 3000
+            required_surplus: 3000
             security_environment: EDC
             ami_id: image_id_ecs_amz2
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-daac.yml
```diff
@@ -23,9 +23,9 @@ jobs:
             deploy_ref: refs/heads/main
             job_files: job_spec/AUTORIFT.yml job_spec/INSAR_GAMMA.yml job_spec/RTC_GAMMA.yml
             instance_types: r6id.xlarge,r5dn.xlarge,r5d.xlarge
-            default_max_vcpus: 1000
-            expanded_max_vcpus: 2000
-            required_surplus: 2000
+            default_max_vcpus: 1500
+            expanded_max_vcpus: 3000
+            required_surplus: 3000
```
asjohnston-asf, line 28: Historically we've kept the `hyp3-edc-uat` environment configured the same as `hyp3-edc-prod` so our testing environment is identical. I'm leaning towards updating the uat values to match, although I'm open to setting quota/vcpus/surplus/budget to values more appropriate to test rather than matching prod. Thoughts?

jhkennedy: I like keeping them exactly the same as well. I suppose we could technically double-spend our budget that way...

asjohnston-asf (approved):

Issue #1272, opened 2022-10-17
Modify `StartExecution` architecture to submit Batch jobs at the maximum rate


jtherrmann:
**Jira:** https://asfdaac.atlassian.net/browse/TOOL-2043

As described by [AWS Batch service quotas](https://docs.aws.amazon.com/batch/latest/userguide/service_limits.html), we're limited to 50 transactions per second (TPS) for each account for [SubmitJob](https://docs.aws.amazon.com/batch/latest/APIReference/API_SubmitJob.html) operations, per AWS Region. So we can submit up to 50 Batch jobs per second.

We refactored our `StartExecution` architecture in https://github.com/ASFHyP3/hyp3/pull/1263 to use the manager/worker model to start batches of step function executions in parallel. Unfortunately this caused executions to be submitted in bursts that overran Batch's `SubmitJob` service limit, so we reduced the amount of concurrency in https://github.com/ASFHyP3/hyp3/pull/1271. 

There are two problems with the current approach:

* Currently our `StartExecution` manager/worker system starts up to 900 executions per minute, or 15 executions per second, which is well under the maximum rate of 50 Batch jobs per second.
* Also, as described [here](https://github.com/ASFHyP3/hyp3/pull/1271#issuecomment-1281246945), there may be some duplicate step function execution attempts which would lead to reduced throughput.

Here is my quick brain dump from 2022-10-14 summarizing my discussion with @asjohnston-asf regarding this issue:

> So Batch is limited to 50 submissions per sec, and we’re starting executions in big bursts that overrun that limit. Our tentative solution is to keep the manager running for 15 min at a time and invoking a steady stream of workers to start jobs at close to the max rate that Batch can handle. We may need to refactor how we get pending jobs; ideally it would operate as a queue so that step function executions are started in the order that the jobs were submitted through the hyp3 API? (Although what if someone submits a ton of jobs right before someone else submits just a few? Will the round-robin priority system take care of that even if the first user’s jobs are all submitted to Batch before the second user’s jobs?)

See [`put_jobs`](https://github.com/ASFHyP3/hyp3/blob/develop/lib/dynamo/dynamo/jobs.py#L36) for the implementation of job priority.

At some point I would like to continue this discussion and improve our `StartExecution` architecture in order to solve the two problems described above.

jtherrmann:
[...]

PR #1365, opened 2022-12-14
draft of limiting number of concurrent step function executions

asjohnston-asf:
FilterExpressions are applied *after* a Query is evaluated, see https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html#Query.Limit . For example:

```python
>>> response = table.query(
...     IndexName='status_code',
...     KeyConditionExpression=Key('status_code').eq('SUCCEEDED'),
...     FilterExpression=Attr('execution_started').eq(False),
... )
>>> pprint.pprint(response)
{'Count': 0,
 'Items': [],
 'LastEvaluatedKey': {'job_id': 'ebeb0696-cab2-4cdd-8d80-be6b2815bb80',
                      'status_code': 'SUCCEEDED'},
 'ResponseMetadata': {'HTTPHeaders': {'connection': 'keep-alive',
                                      'content-length': '148',
                                      'content-type': 'application/x-amz-json-1.0',
                                      'date': 'Wed, 14 Dec 2022 21:02:34 GMT',
                                      'server': 'Server',
                                      'x-amz-crc32': '3368818628',
                                      'x-amzn-requestid': 'KSAE99641TFOAUKOIL7PC9LJEJVV4KQNSO5AEMVJF66Q9ASUAAJG'},
                      'HTTPStatusCode': 200,
                      'RequestId': 'KSAE99641TFOAUKOIL7PC9LJEJVV4KQNSO5AEMVJF66Q9ASUAAJG',
                      'RetryAttempts': 0},
 'ScannedCount': 766}
 ```

We already have a natural limit on the number of concurrent step function executions, as `get_jobs_waiting_for_execution` has to scan through more and more jobs where `execution_started=True` before it finds 900 where `execution_started=False`. Eventually the StartExecutionManager lambda hits its 10s timeout while the query is running and no new step function executions are started. Anecdotally, the tipping point is in the hundreds of thousands of executions.

This PR makes that limit visible and lower. This change should have no impact on throughput; as in-progress executions complete new executions will be submitted at the same rate; up to our maximum of 900 per minute.

Pros:

As more and more step functions are running concurrently, we see more and more batch.DescribeJob calls made as each of those step functions polls for the status of its AWS Batch job. This generates overhead costs from AWS GuardDuty in JPL/EDC environments. It also leads to `TooManyRequestsException` errors for other individuals/apps making batch.DescribeJob calls. Reducing the number of concurrent executions reduces the impact of these issues.

A lower concurrency limit better positions us to pause major processing campaigns and release code fixes if/when needed, since fewer jobs will be "committed" as a running step function execution.

Cons:

This impacts our job priority scheme if/when the concurrency limit is reached. Job priority is implemented at the Batch job level. StartExecution processes jobs independent of priority (I'm not even sure it's first-come first-served). When HyP3 is at the concurrency limit, new high-priority jobs would have to wait in line until StartExecution processes them, at which they'd jump to the front of the 50,000 in-progress executions.

HyP3 production is at little risk of hitting any concurrency limit, since we typically process < 100,000 jobs per month.

### DIFF
--- a/lib/dynamo/dynamo/jobs.py
+++ b/lib/dynamo/dynamo/jobs.py
@@ -168,5 +168,6 @@ def get_jobs_waiting_for_execution(limit: int, max_scanned: int = 50000) -> list
         params['ExclusiveStartKey'] = response['LastEvaluatedKey']
         response = table.query(**params)
         jobs.extend(response['Items'])
+        scanned += response['ScannedCount']
 
     return jobs[:limit]
--- a/lib/dynamo/dynamo/jobs.py
+++ b/lib/dynamo/dynamo/jobs.py
@@ -152,7 +152,7 @@ def update_job(job):
     )
 
 
-def get_jobs_waiting_for_execution(limit: int, max_scanned: int = 50000) -> list[dict]:
+def get_jobs_waiting_for_execution(limit: int, max_scanned: int = 50_000) -> list[dict]:
     table = DYNAMODB_RESOURCE.Table(environ['JOBS_TABLE_NAME'])
 
     params = {
--- a/lib/dynamo/dynamo/jobs.py
+++ b/lib/dynamo/dynamo/jobs.py
@@ -152,7 +152,7 @@ def update_job(job):
     )
 
 
-def get_jobs_waiting_for_execution(limit: int) -> list[dict]:
+def get_jobs_waiting_for_execution(limit: int, max_scanned: int = 50000) -> list[dict]:
     table = DYNAMODB_RESOURCE.Table(environ['JOBS_TABLE_NAME'])
 
     params = {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/dynamo/dynamo/jobs.py [hunk 1]
```diff
@@ -152,7 +152,7 @@ def update_job(job):
     )
 
 
-def get_jobs_waiting_for_execution(limit: int) -> list[dict]:
+def get_jobs_waiting_for_execution(limit: int, max_scanned: int = 50000) -> list[dict]:
```
jtherrmann, line 155:
```suggestion
def get_jobs_waiting_for_execution(limit: int, max_scanned: int = 50_000) -> list[dict]:
```

asjohnston-asf:
Discussing the implementation details with @jtherrmann, the business logic behind this implementation is very inside-baseball and relies on DynamoDB returning query results in a consistent order. We'll put more thought into considering alternatives before trying to push a change through.

jtherrmann:
We may want to move the broader discussion to https://github.com/ASFHyP3/hyp3/issues/1272

PR #1391, opened 2022-12-31
set OMP_NUM_THREADS for GAMMA job types and expand DAAC EC2 instance types

asjohnston-asf:
Previously, running multiple GAMMA jobs on the same EC2 instance led to severe CPU contention. For example, RTC_GAMMA jobs typically finish in 15-20m when running a single job on an `r5d.xlarge` instance. Running four RTC_GAMMA jobs in parallel on an `r5d.4xlarge` instance would cause most jobs to not finish before their 90 minute timeout.

With this change, running four RTC_GAMMA jobs on a single `r5d.4xlarge` instance is marginally *faster* than running them on four separate `r5d.xlarge` instances. Similar results for INSAR_GAMMA jobs.

This is a significant change that will require monitoring in production for any unintended cost or performance changes.

I've added all memory-optimized instances that match the `r5d.xlarge` proportions of 4 vCPU and 150 GB disk per  32 GB memory to the DAAC deployments:
- `r6id.xlarge`
- `r6id.2xlarge`
- `r6id.4xlarge`
- `r6id.8xlarge`
- `r5dn.xlarge`
- `r5dn.2xlarge`
- `r5d.xlarge`
- `r5d.2xlarge`

Larger instance types in each family break their disk space into multiple SSDs. Docker can only leverage on of those drives, so these larger instance types don't have sufficient disk to support being fully loaded with 10x2 InSAR jobs.

### DIFF
--- a/.github/workflows/deploy-daac.yml
+++ b/.github/workflows/deploy-daac.yml
@@ -22,7 +22,7 @@ jobs:
             quota: 1000
             deploy_ref: refs/heads/main
             job_files: job_spec/AUTORIFT.yml job_spec/INSAR_GAMMA.yml job_spec/RTC_GAMMA.yml
-            instance_types: r6id.xlarge,r5dn.xlarge,r5d.xlarge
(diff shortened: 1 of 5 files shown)


asjohnston-asf:
Requires https://github.com/ASFHyP3/hyp3-gamma/pull/428

jacquelynsmale (approved):

PR #1410, opened 2023-01-10
increase RTC_GAMMA timeout from 3 hours to 10 hours

asjohnston-asf:


forrestfwilliams (approved):
Looks good, 10 hrs seems to be a reasonable upper limit.

PR #1439, opened 2023-01-23
Add details on step function updates to release checklist

jtherrmann:


.github/workflows/release-template-comment.yml
```diff
@@ -26,7 +26,12 @@ jobs:
             - [ ] (If applicable) Updated the dependencies and indicated any downstream changes that are required
             - [ ] Added/updated documentation for these changes
             - [ ] Added/updated tests for these changes
-            - [ ] If the step function code has changed, have you drained the job queue before merging?
+            - [ ] If the step function code has changed\*, have you drained the job queue before merging?
+
+            \* For example, if the interface for a Lambda function has changed to expect different input,
+            then currently running jobs (which use the old step function definition) will call the new
+            function with the old input. So we must drain the job queue before deployment, so that the new
+            function is only called by the new step function definition.
```
jhkennedy, line 34:
```suggestion
            - [ ] If the step function code has changed, have you drained the job queue before merging?

              * For example, if the interface for a Lambda function has changed to expect different input,
                then currently running jobs (which use the old step function definition) will call the new
                function with the old input. So we must drain the job queue before deployment, so that the new
                function is only called by the new step function definition.
```

Instead of foot-note style, I'd prefer this to be a sub-item. See comparison below

--- 

template from PR shows up like so:

---

### Developer checklist
- [ ] Indicated the level of changes to this package by affixing one of these labels:
  * major -- Major changes to the API that may break current workflows
  * minor -- Minor changes to the API that do not break current workflows
  * patch -- Patches and bugfixes for the current version that do not break current workflows
  * bumpless -- Changes to documentation, CI/CD pipelines, etc. that don't affect the software's version
- [ ] (If applicable) Updated the dependencies and indicated any downstream changes that are required
- [ ] Added/updated documentation for these changes
- [ ] Added/updated tests for these changes
- [ ] If the step function code has changed\*, have you drained the job queue before merging?

\* For example, if the interface for a Lambda function has changed to expect different input,
then currently running jobs (which use the old step function definition) will call the new
function with the old input. So we must drain the job queue before deployment, so that the new
function is only called by the new step function definition.

### Reviewer checklist

---

vs

---

### Developer checklist
- [ ] Indicated the level of changes to this package by affixing one of these labels:
  * major -- Major changes to the API that may break current workflows
  * minor -- Minor changes to the API that do not break current workflows
  * patch -- Patches and bugfixes for the current version that do not break current workflows
  * bumpless -- Changes to documentation, CI/CD pipelines, etc. that don't affect the software's version
- [ ] (If applicable) Updated the dependencies and indicated any downstream changes that are required
- [ ] Added/updated documentation for these changes
- [ ] Added/updated tests for these changes
- [ ] If the step function code has changed, have you drained the job queue before merging?
  * For example, if the interface for a Lambda function has changed to expect different input,
    then currently running jobs (which use the old step function definition) will call the new
    function with the old input. So we must drain the job queue before deployment, so that the new
    function is only called by the new step function definition.

### Reviewer checklist

---

jhkennedy (dismissed):
Looks good to me! One sylistic comment feel free to take or leave.

jtherrmann:
> Looks good to me! One sylistic comment feel free to take or leave.

I like it.

jhkennedy (approved):

PR #1460, opened 2023-01-31
Enable ESD in INSAR_ISCE_TEST

### LINKED ISSUE #1448: Option to enable Enhanced Spectral Diversity correction in INSAR_ISCE output
[DockerizedTopsApp](https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp) has exposed an option to add Enhanced Spectral Diversity (ESD) correction when creating the GUNW product. 

Enabling/disabling ESD corrections is done via the `--do-esd` parameter, which takes `true`/`false` values, to the DockerizedTopsApp container HyP3 entry point. See ACCESS-Cloud-Based-InSAR/DockerizedTopsApp#74 

Additionally, there is an `--esd-coherence-threshold` parameter which accepts `float` values between 0 and 1. 

This feature request is to:
*  add a new `do_esd` parameter to the `INSAR_ISCE` and `INSAR_ISCE_TEST` jobs with possible values of `true` and `false`
*  add a new `esd_coherence_threshold` parameter to `INSAR_ISCE` and `INSAR_ISCE_TEST` jobs which accepts `float` values between 0 and 1 (inclusively). We should set a default value of 0.85 for this parameter.
E.g., 

```json
{
  "job_type": "INSAR_ISCE",
  "job_parameters": {
    "granules": ["S1B_IW_SLC__1SDV_20210723T014947_20210723T015014_027915_0354B4_B3A9"],
    "secondary_granules": ["S1B_IW_SLC__1SDV_20210711T014947_20210711T015013_027740_034F80_D404"],
    "do_esd": "true",
    "esd_coherence_threshold": 0.5
  }
}
```
A job requested with `"do_eds": "false"` should produce the same output as currently produced. If `true` is specified, the output NetCDF product should ???

forrestfwilliams:
first part of #1448

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.1.0]
+### Added
+- `INSAR_ISCE_TEST` jobs now accept  `do_esd`, and `esd_coherence_threshold` parameters to specify whether or not to perform the Enhanced Spectral Diversity (ESD), and what ESD coherence threshold to use.
+
 ## [3.0.0]
 ### Added
 - Added `hyp3-bgc-engineering` deployment.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [3.1.0]
 ### Added
-- `INSAR_ISCE_TEST` jobs now accept  `do_esd`, and `esd_coherence_threshold` parameters to specify whether or not to perform the Enhanced Spectral Diversity (ESD), and what ESD coherence threshold to use.
+- `INSAR_ISCE_TEST` jobs now accepts an `esd_coherence_threshold` parameter to specify whether or not to perform the Enhanced Spectral Diversity (ESD), and what ESD coherence threshold to use.
 
 ## [3.0.0]
 ### Added
--- a/job_spec/INSAR_ISCE_TEST.yml
+++ b/job_spec/INSAR_ISCE_TEST.yml
@@ -42,7 +42,7 @@ INSAR_ISCE_TEST:
         description: Whether or not to perform Enhanced Spectral Diversity for improved coregistration
         defult: false
         type: boolean
-    esd-coherence-threshold:
+    esd_coherence_threshold:
       api_schema:
         description: Coherence threshold for points to include when performing ESD
         defult: 0.5
--- a/job_spec/INSAR_ISCE_TEST.yml
+++ b/job_spec/INSAR_ISCE_TEST.yml
@@ -73,6 +73,8 @@ INSAR_ISCE_TEST:
         - Ref::granules
         - --secondary-scenes
         - Ref::secondary_granules
+        - --esd-coherence-threshold
+        - Ref::esd_coherence_threshold
         - --estimate-ionosphere-delay
         - Ref::estimate_ionosphere_delay
         - --frame-id
--- a/job_spec/INSAR_ISCE_TEST.yml
+++ b/job_spec/INSAR_ISCE_TEST.yml
@@ -47,8 +47,8 @@ INSAR_ISCE_TEST:
         description: Coherence threshold for points to include when performing ESD
         defult: 0.5
         type: number
-        min: 0.0
-        max: 1.0
+        minimum: 0.0
+        maximum: 1.0
     bucket_prefix:
       default:  '""'
   validators: []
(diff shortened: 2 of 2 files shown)


job_spec/INSAR_ISCE_TEST.yml
```diff
@@ -37,6 +37,18 @@ INSAR_ISCE_TEST:
...
           - GMAO
+    do_esd:
+      api_schema:
+        description: Whether or not to perform Enhanced Spectral Diversity for improved coregistration
+        defult: false
+        type: boolean
+    esd_coherence_threshold:
+      api_schema:
+        description: Coherence threshold for points to include when performing ESD
+        defult: 0.5
```
jhkennedy, line 48:
```suggestion
        defult: 0.85
```
To reflect default in ISCE2: https://github.com/isce-framework/isce2/blob/main/applications/topsApp.py#L238

jhkennedy:
I've unlinked #1448 as we don't want to close it until we've also enabled it for `INSAR_ISCE` jobs.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.1.0]
+### Added
+- `INSAR_ISCE_TEST` jobs now accept  `do_esd`, and `esd_coherence_threshold` parameters to specify whether or not to perform the Enhanced Spectral Diversity (ESD), and what ESD coherence threshold to use.
```
jhkennedy, line 9:
```suggestion
- `INSAR_ISCE_TEST` jobs now accept an `esd_coherence_threshold` parameters to specify whether or not to perform the Enhanced Spectral Diversity (ESD), and what ESD coherence threshold to use.
```

job_spec/INSAR_ISCE_TEST.yml
```diff
@@ -37,6 +37,18 @@ INSAR_ISCE_TEST:
...
+        type: number
+        minimum: 0.0
+        maximum: 1.0
```
jhkennedy, line 51:
```suggestion
    esd_coherence_threshold:
      api_schema:
        description: Coherence threshold value, between 0 and 1, for points to include when performing ESD, or -1 for no ESD correction.
        defult: 0.5
        type: number
        minimum: -1.0
        maximum: 1.0
```
Technically, we'll accept values between -1 and 0 which will fail quickly as the container should raise a `ValueError` in that case. I don't think there's a way to say "either -1, or something between 0 and 1" here.

jhkennedy (commented):
Based on the discussion in #1448, and plugin changes in https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/pull/96,  we can drop the `--do-esd` parameter and drive entirely off of the threshold.

job_spec/INSAR_ISCE_TEST.yml
```diff
@@ -37,6 +37,13 @@ INSAR_ISCE_TEST:
           - HRRR
           - HRES
           - GMAO
+    esd_coherence_threshold:
+      api_schema:
+        description: Coherence threshold value, between 0 and 1, for points to include when performing ESD, or -1 for no ESD correction.
+        default: -1.0
+        type: number
+        minimum: -1.0
+        maximum: 1.0
```
jhkennedy, line 46: Technically, this should be a range of `0.` to `1.`, OR `-1.`, but I don't think OpenAPI spec can do that, nor do I think it's necessarily worth doing. @asjohnston-asf @jtherrmann do you know by chance?

jhkennedy (approved):
Just one curiosity that's probably not even worth exploring. :shipit: anytime!

job_spec/INSAR_ISCE_TEST.yml
```diff
@@ -60,7 +60,7 @@ INSAR_ISCE_TEST:
         - Ref::granules
         - --secondary-scenes
         - Ref::secondary_granules
-      timeout: 10800
+      timeout: 21600
```
jhkennedy, line 63: According to @cmarshak ESD (and ionosphere) may 2x (or more) the job run time. This doubles the timout

job_spec/INSAR_ISCE_TEST.yml
jtherrmann: Not seeing a way to do that: https://swagger.io/docs/specification/data-models/data-types/#numbers

job_spec/INSAR_ISCE_TEST.yml
```diff
@@ -27,6 +27,13 @@ INSAR_ISCE_TEST:
           minLength: 67
           maxLength: 67
           example: S1B_IW_SLC__1SDV_20210711T014947_20210711T015013_027740_034F80_D404
+    esd_coherence_threshold:
+      api_schema:
+        description: Coherence threshold value, between 0 and 1, for points to include when performing ESD, or -1 for no ESD correction.
```
jtherrmann, line 32:
```suggestion
        description: Coherence threshold value for determining which points to include when performing ESD. Must be any value between 0 and 1 (inclusive), or -1 for no ESD correction.
```

forrestfwilliams: Yep, I prefer this wording. Thanks @jtherrmann

jtherrmann (approved):

PR #1490, opened 2023-02-11
Fix OpenAPI validator

jtherrmann:
TODO:

- [ ] Fix the tests that are still failing
- [ ] Some of the subscription tests were failing because the mock subscriptions didn't include all of the required parameters. I'm not sure why we weren't seeing these errors before? Might be worth investigating further.


jtherrmann:
This will be replaced by https://github.com/ASFHyP3/hyp3/pull/1714

PR #1510, opened 2023-02-23
Proposal on how to add additional secrets for job tasks

jhkennedy:
Right now, we provide secrets to plugins (containers) via their respective Batch job definition, and every Batch job requires the same credentials (Earthdata login username and password).

However, the troposphere task in the INSAR_ISCE_TEST job requires additional secrets to be passed to the RAiDER plugin.

---

This PR does two major things:
* Job tasks may now specify the names of any needed secrets in a `secrets` list in their `job_spec` task definition
* The Secret in AWS Secrets containing the needed secrets *will now be externally managed* meaning it will need to be set up and managed outside of the HyP3 cloud formation stack (I would expect we'd create the Secret and populate it as part of the initial deployment like we do for the SSL Certificate in the AWS Certificate Manager and then modify it in the AWS console as necessary)
  * the `EDLUsername` and `EDLPassword` cloud formation parameters have been replaced with a `SecretArn` parameter
  * adjusts the deploy action and workflows accordingly
    * a new `SECRET_ARN` secret needs to be in the GitHub Deploy Environment
    * `EDL_USERNAME` and `EDL_PASSWORD` secrets can be dropped from the GitHub Deploy Environment if they exist

---

## TODOs

To get this into test we'll need to:
- [x] Manually provision a Secrets Manager Secret with all the [needed secrets](https://github.com/ASFHyP3/hyp3/blob/tropo-env-vars/job_spec/INSAR_ISCE_TEST.yml#L110-L116)
      * RAiDER secrets have been shared with us in the JPL A19 AWS account
- [x] Add the manually provisioned Secret ARN to the hyp3-enterprise-test and hyp3-test (UAT) environment
- [x] Update README accordingly
- [x] Update deployment docs accordingly


apps/workflow-cf.yml.j2
```diff
@@ -72,6 +72,10 @@ Resources:
             ValueFrom: !Sub "${SecretArn}:username::"
           - Name: EARTHDATA_PASSWORD
             ValueFrom: !Sub "${SecretArn}:password::"
```
jhkennedy, line 74: Instead of *always* providing these, I'm tempted to move them into the `job_spec` task definitions too.

forrestfwilliams: I think I'm leaning towards adding these to the `job_spec` task definitions. That way it's very explicit what secrets each plugin is using.

job_spec/WATER_MAP.yml
```diff
@@ -175,3 +181,6 @@ WATER_MAP:
       timeout: 86400
       vcpu: 1
       memory: 126000
+      secrets:
+        - EARTHDATA_USERNAME
+        - EARTHDATA_PASSWORD
```
jhkennedy, line 186: :thinking: I'm not actually sure if `asf_tools` needs EDL credentials... will need to check.

job_spec/WATER_MAP.yml
```diff
@@ -153,6 +156,9 @@ WATER_MAP:
       timeout: 36000
       vcpu: 1
       memory: 126000
+      secrets:
+        - EARTHDATA_USERNAME
+        - EARTHDATA_PASSWORD
```
jhkennedy, line 161: :thinking: I'm not actually sure if `asf_tools` needs EDL credentials... will need to check.

job_spec/RIVER_WIDTH.yml
```diff
@@ -84,3 +90,6 @@ RIVER_WIDTH:
       timeout: 36000
       vcpu: 1
       memory: 31600
+      secrets:
+        - EARTHDATA_USERNAME
+        - EARTHDATA_PASSWORD
```
jhkennedy, line 95: :thinking: I'm not actually sure if `RiverWidthSAR` needs EDL credentials... will need to check.

job_spec/WATER_MAP.yml
jhkennedy: It does not.

job_spec/WATER_MAP.yml
jhkennedy: It does not.

job_spec/RIVER_WIDTH.yml
```diff
@@ -74,6 +77,9 @@ RIVER_WIDTH:
       timeout: 36000
       vcpu: 1
       memory: 126000
+      secrets:
+        - EARTHDATA_USERNAME
+        - EARTHDATA_PASSWORD
```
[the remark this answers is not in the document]
jhkennedy: It does not.

apps/workflow-cf.yml.j2
```diff
@@ -67,11 +67,13 @@ Resources:
           {% for command in task['command'] %}
           - {{ command }}
           {% endfor %}
+        {% if task.get('secrets', []) %}
```
jtherrmann, line 70: Given that `None` and `[]` are both falsy, I don't think we need to specify `[]` as the default value here?
```suggestion
        {% if task.get('secrets') %}
```

jtherrmann, line 70: I believe you could get rid of the `if` block entirely, and edit the `for` statement to be: `for secret in task.get('secrets', [])`

jtherrmann, line 70: I don't think we need to specify `[]` as the default value here, given that `None` and `[]` are both falsy.
```suggestion
        {% if task.get('secrets') %}
```

jtherrmann (approved):

jhkennedy:
Alright, everything should be up to date pending README/docs feedback. 

I've created a Secrets Manager Secret and  updated the Action environment for the  `hyp3-edc-uat` and `hyp3-enterprise-deployments, so we should be able to merge this and test RAiDER for HRES and ERA5

docs/deployments/ASF-deployment.md
```diff
@@ -24,12 +24,14 @@ suitable bucket if you try and create a new CloudFormation Stack in the AWS Cons
...
-  integrate the deployment into our CI/CD pipelines
+1. Set the account-wide API Gateway logging permissions
+2. A deployment role with the necessary permissions to deploy HyP3
+3. A "service user" so that we can generate long-term AWS access keys and
+   integrate the deployment into our CI/CD pipelines
+4. An AWS Secrets Manager Secret containing key-value pairs for all secrets listed in the [`job_spec`s](./job_spec)
+   which will be deployed
 
-These can be done by deploying the `hyp3-ci` stack. From the repository root, run:
+(1) thorugh (3) can be done by deploying the `hyp3-ci` stack. From the repository root, run:
```
jtherrmann, line 34:
```suggestion
(1) through (3) can be done by deploying the `hyp3-ci` stack. From the repository root, run:
```

docs/deployments/ASF-deployment.md
```diff
@@ -78,5 +93,6 @@ aws cloudformation deploy \
         EDLUsername=<earthdata-login-username> \
```
jtherrmann, line 93: Should we remove `EDLUsername` and `EDLPassword`?

jhkennedy: Ah rats, I thought I did! Fixed.

jtherrmann (approved):

PR #1516, opened 2023-02-27
Release 3.1.2 -- SET for INSAR_ISCE_TEST

### LINKED ISSUE #1447: Option to enable Solid Earth Tides correction in INSAR_ISCE_TEST output
[DockerizedTopsApp](https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp) will contain tools to calculate Solid Earth Tide (SET) corrections and add them as additional layers to the GUNW product. 

Enabling/disabling adding SET corrections layers is done via the `--compute-solid-earth-tide` parameter, which takes `true`/`false` values, to the DockerizedTopsApp container HyP3 entry point. See ACCESS-Cloud-Based-InSAR/DockerizedTopsApp#91 

This feature request is to add a new `compute_solid_earth_tide` parameter to the `INSAR_ISCE` and `INSAR_ISCE_TEST` jobs with possible values of `true` and `false`. E.g., 

```json
{
  "job_type": "INSAR_ISCE",
  "job_parameters": {
    "granules": ["S1B_IW_SLC__1SDV_20210723T014947_20210723T015014_027915_0354B4_B3A9"],
    "secondary_granules": ["S1B_IW_SLC__1SDV_20210711T014947_20210711T015013_027740_034F80_D404"],
    "compute_solid_earth_tide": "true"
  }
}
```
A job requested with `"compute_solid_earth_tide": "false"` should produce the same output as currently provided. If `true` is specified, the output NetCDF product should be the output including the new SET data layer:

```
/science/grids/corrections/external/tides/solidEarthTide
```

jhkennedy:
Fixes #1447

### DIFF
--- a/.github/workflows/deploy-daac.yml
+++ b/.github/workflows/deploy-daac.yml
@@ -44,7 +44,7 @@ jobs:
               job_spec/AUTORIFT.yml
               job_spec/INSAR_GAMMA.yml
               job_spec/RTC_GAMMA.yml
-              job_spec/INSAR_ISCE.yml
+              job_spec/INSAR_ISCE_TEST.yml
               job_spec/WATER_MAP.yml
             instance_types: r6id.xlarge,r6id.2xlarge,r6id.4xlarge,r6id.8xlarge,r5dn.xlarge,r5dn.2xlarge,r5d.xlarge,r5d.2xlarge
             default_max_vcpus: 1500
--- a/.github/workflows/deploy-daac.yml
+++ b/.github/workflows/deploy-daac.yml
@@ -103,6 +103,6 @@ jobs:
   call-bump-version-workflow:
     if: github.ref == 'refs/heads/main'
     needs: deploy
-    uses: ASFHyP3/actions/.github/workflows/reusable-bump-version.yml@v0.4.0
+    uses: ASFHyP3/actions/.github/workflows/reusable-bump-version.yml@v0.7.0
     secrets:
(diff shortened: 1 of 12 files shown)


.github/workflows/deploy-enterprise-test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         include:
           - environment: hyp3-enterprise-test
             domain: hyp3-enterprise-test.asf.alaska.edu
-            template_bucket: cf-templates-aubvn3i9olmk-us-west-2
+            template_bucket: cf-templates-1iw894v4yzqya-us-west-2
```
forrestfwilliams, line 19: Is there a reason that this bucket changed? Is it auto-generated?

.github/workflows/deploy-daac.yml
```diff
@@ -44,7 +44,7 @@ jobs:
               job_spec/AUTORIFT.yml
               job_spec/INSAR_GAMMA.yml
               job_spec/RTC_GAMMA.yml
-              job_spec/INSAR_ISCE.yml
+              job_spec/INSAR_ISCE_TEST.yml
```
forrestfwilliams, line 47: It looks like this change is confined to the EDC-UAT deployment, but I just want to double-check with @jhkennedy that this change won't impact any of our other deployments

forrestfwilliams (approved):

PR #1582, opened 2023-04-20
Increase INSAR_ISCE_TEST DockerizedTopsApp timeout

jhkennedy:
https://hyp3-test-api.asf.alaska.edu/jobs/f258c8a3-6763-49e6-afae-8687cdb7aed6 appears to have failed to to hitting the timeout:
image

We previously 2x the production INSAR_ISCE DockerizedTopsApp step timeout, going from 10800 to 21600. This puts it at 32400 (3x).

@cmarshak thoughts on this timeout?

### DIFF
--- a/job_spec/INSAR_ISCE_TEST.yml
+++ b/job_spec/INSAR_ISCE_TEST.yml
@@ -85,7 +85,7 @@ INSAR_ISCE_TEST:
         - Ref::frame_id
         - --compute-solid-earth-tide
         - Ref::compute_solid_earth_tide
-      timeout: 21600
+      timeout: 32400
       vcpu: 1
       memory: 7500
       secrets:


jhkennedy:
Oh actually, this could just be resource contention if multiple jobs were run at the same time on the same instance because of ACCESS-Cloud-Based-InSAR/DockerizedTopsApp#106...

jhkennedy:
Soo looking at this, 2 jobs all finished right around the same time, 1 succeeded at 5 hours 56-ish minutes, and 1 failed hitting the 6 hour timeout. It's likely they both ran on the same instance due to the instance types available in this deployment:
https://github.com/ASFHyP3/hyp3/blob/develop/.github/workflows/deploy-daac.yml#L49
And the resource requirements needed:
https://github.com/ASFHyP3/hyp3/blob/develop/job_spec/INSAR_ISCE_TEST.yml#L89-L90
which would allow packing of multiple jobs.

Since one did succeed, upping the timeout *might* help, but I think addressing ACCESS-Cloud-Based-InSAR/DockerizedTopsApp#106 should also be done first and might solve it.

cmarshak:
What are the job parameters? SLC IDs?

If you are using ESD, that's very much untested...

With ionosphere, we are still just testing these using frames, which may be spatially more constricted.

There is a lot of ways this could be hitting the timeout. I haven't seen the timeout in my testing yet.

cmarshak:
All I see in the link is: 

```
{"detail":"No authorization token provided","status":401,"title":"Unauthorized","type":"about:blank"}
```

jhkennedy:
> {"detail":"No authorization token provided","status":401,"title":"Unauthorized","type":"about:blank"}

@cmarshak that means your ASF cookie expired and you just need to log in to vertex <search.asf.alaska.edu> -- that can/will happen on any of the JPL deployments as well.

cmarshak:
Thanks, Joe! I see it now:

```
{"execution_started":true,"expiration_time":"2023-05-02T00:00:00+00:00","job_id":"f258c8a3-6763-49e6-afae-8687cdb7aed6","job_parameters":{"compute_solid_earth_tide":false,"esd_coherence_threshold":-1,"estimate_ionosphere_delay":true,"frame_id":-1,"granules":["S1A_IW_SLC__1SDV_20230401T100922_20230401T100949_047901_05C183_A3C1"],"secondary_granules":["S1A_IW_SLC__1SDV_20221120T100925_20221120T100952_045976_05806E_F4CC"],"weather_model":"None"},"job_type":"INSAR_ISCE_TEST","logs":["https://d1riv60tezqha9.cloudfront.net/f258c8a3-6763-49e6-afae-8687cdb7aed6/f258c8a3-6763-49e6-afae-8687cdb7aed6.log"],"name":"isce_ion_test","priority":9997,"processing_times":[21634.385],"request_time":"2023-04-16T18:19:03+00:00","status_code":"FAILED","user_id":"scottyhq"}
```

I honestly would have to run it locally and see. Not sure... It's 1 reference and 1 secondary so should be doable. I have been running using frames. See the notebook above how I enumerate and submit.

I am seeing about 3.5 hours for ionosophere turned on: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/discussions/94#discussioncomment-4976998

I have seen lots of transient errors in job submissions. I often resubmit and it goes away.

Also, ISCE2 (not known for its friendly logging/error handling) happens to print out quite a bit to the standard io. As a result, when something fails due to a long running process, there is so much to capture, hyp3 + aws batch has a hard time getting the exact point things are ending. Even if you log into aws, it's still hard to figure out what's happening in batch (the logs have to be filtered to see anything).

PR #1589, opened 2023-04-27
Add 20m watermap jobs

jtherrmann:


apps/step-function.json.j2
asjohnston-asf, line 1: The water map memory settings look like they'll do what we want, just mentioning here that the workflow is a little awkward. The default memory for the water maps steps in WATER_MAP.yml are:
```
rtc - 31600
water map - 126000
flood map - 126000
```
The 10/20m switch ultimately sets the rtc memory to 126000, even though 10m rtc jobs only need 63200, because we can't currently override on a task-by-task basis.

asjohnston-asf (dismissed):

apps/step-function.json.j2
jtherrmann: Agreed, this is not ideal.

asjohnston-asf (approved):

jtherrmann:
Test job: https://hyp3-test-api.asf.alaska.edu/jobs/8ac5041c-9e8b-4ea4-8cfe-3d4b971cd15c

Verify:

- [ ] Memory was set as expected.
- [ ] Output product is 20m and looks as expected.

PR #1606, opened 2023-05-04
Release v3.5.0

asjohnston-asf:


asjohnston-asf:
I don't believe the step function change poses any risk to running jobs, since it isn't connected to any python code changes.

jtherrmann (approved):
I want to hold off on merging until we confirm that memory is being set correctly for `WATER_MAP_EQ` 10m jobs.

jtherrmann (approved):

Issue #1273, opened 2022-10-17
Include job URLs in `post_jobs` response


jtherrmann:
I want the response for the hyp3 API's [`post_jobs`](https://hyp3-test-api.asf.alaska.edu/ui/#/default/post_jobs) endpoint to include the API URL for each job, e.g. `https://hyp3-test-api.asf.alaska.edu/jobs/4bcb674a-02b1-49d4-98e0-e97aa3249067`. This could be a field of the job dict but it would be even more convenient to have a `job_urls` field as the first key in the response dict with a corresponding value which is a list of the job URLs. Then, when submitting a batch of sample jobs, I could just download the response, open the JSON file in my browser (which conveniently detects URLs in JSON and displays them as hyperlinks), and click each link to open all of the jobs, or just copy-paste the list of URLs to somewhere that other members of the team can view it.

Even more convenient would be if we could get the Swagger UI to include a button to open the JSON response in the browser rather than having to first download it and then open it.

jtherrmann:
And [here](https://github.com/ASFHyP3/hyp3/blob/develop/apps/api/src/hyp3_api/routes.py#L136) is where the response gets serialized to JSON, which would be where we determine the dict key order to ensure that `job_urls` is the first key.

PR #1616, opened 2023-05-10
Add job_urls field to post_jobs response

jtherrmann:
See https://github.com/ASFHyP3/hyp3/issues/1273

This PR adds a `job_urls` field to the `POST /jobs` response, so that we do not have to manually concatenate the API domain name and the job ID for every job when submitting jobs via the API. Here is an example response:

```json
{
  "job_urls": [
    "https://hyp3-enterprise-test.asf.alaska.edu/jobs/13d6eac4-b4e6-4281-951e-d79d356693bf",
    "https://hyp3-enterprise-test.asf.alaska.edu/jobs/1693d659-6bba-495e-9332-b68dcd753e57",
    "https://hyp3-enterprise-test.asf.alaska.edu/jobs/ac2c7a9f-d278-4673-89b1-bf7faa7118c3"
  ],
  "jobs": [...],
  "validate_only": false
}
```

This makes it much easier to open and keep track of a batch of submitted test jobs, for example, or to comment the list of URLs in an open task/issue.

TODO:

- [ ] Address `TODO`s in code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,8 +114,8 @@ The API can be run locally to verify changes, but must be connected to a set of 
 1. Set up AWS credentials in your environment as described
    [here](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration).
    Also see our [wiki page](https://github.com/ASFHyP3/.github-private/wiki/AWS-Access#aws-access-keys).
-2. Edit `tests/cfg.env` to specify the names of existing DynamoDB tables from a particular HyP3 deployment,
-   as well as the corresponding API domain name. Delete all of the `AWS_*` variables.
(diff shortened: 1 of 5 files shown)


jhkennedy:
@jtherrmann ooh, this could be handy. 

I wonder if we should also (instead?) add a `job_url` to each job -- when you do searches for multiple jobs (e.g., by name) you'd have URLs to open that specific job.

jtherrmann:
@jhkennedy I think I'd like both. My only dilemma with adding a `job_url` field to each job dict is whether to store it as a field in the database, or insert it dynamically when returning the job via the API.

jtherrmann:
I don't have as much need for this feature since we added the search-for-other-users'-jobs feature.

PR #1617, opened 2023-05-11
increase memory available to INSAR_GAMMA jobs in azdwr-hyp3 deployment

asjohnston-asf:
Change from INSAR_GAMMA.yml:
- increased memory to 63200
- increased omp-num-threads to 8

```
$ diff INSAR_GAMMA.yml INSAR_GAMMA_HIGH_MEMORY.yml 
75c75
<         - '4'
---
>         - '8'
99c99
<       memory: 31600
---
>       memory: 63200
```

### DIFF
--- a/job_spec/INSAR_GAMMA_HIGH_MEMORY.yml
+++ b/job_spec/INSAR_GAMMA_HIGH_MEMORY.yml
@@ -1,0 +1,102 @@
+INSAR_GAMMA:
+  required_parameters:
+    - granules
+  parameters:
+    granules:
+      default:  '""'
+      api_schema:
+        type: array
+        minItems: 2
+        maxItems: 2
+        example:
+          - S1A_IW_SLC__1SSV_20150621T120220_20150621T120232_006471_008934_72D8
+          - S1A_IW_SLC__1SSV_20150504T120217_20150504T120229_005771_00769E_EF9A
+        items:
+          description: The name of the Sentinel-1 SLC granule to process
+          type: string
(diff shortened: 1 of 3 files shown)


job_spec/INSAR_GAMMA_HIGH_MEMORY.yml
```diff
@@ -0,0 +1,102 @@
+INSAR_GAMMA:
```
jtherrmann, line 1: Maybe? I don't actually know if this should change.
```suggestion
INSAR_GAMMA_HIGH_MEMORY:
```

jtherrmann, line 1: Maybe? I don't actually know if this should change.
```suggestion
INSAR_GAMMA_HIGH_MEMORY:
```

asjohnston-asf: The job_type you give to the API is driven by the top-level dictionary key, not the file name. This lets azdwr-hyp3 continue to use the same job name (`INSAR_GAMMA`) but get the higher memory reservation.

We do the same thing with AUTORIFT.yml and AUTORIFT_ITS_LIVE.yml, so jobs in both edc-prod and hyp3-autorift are named `AUTORIFT`, but are run with different configurations.

jtherrmann: Cool, thanks for clearing that up.

jtherrmann (approved):

PR #1640, opened 2023-05-23
Make flood-depth-estimator nullable 

jacquelynsmale:
Started the ball rolling on how to make the `flood-depth-estimator` value functionally nullable in `WATER_MAP` and `WATER_MAP_TEST`. The current implementation will not accept `null` as valid and gets   `"status": 400,
  "title": "Bad Request"`. 
@jhkennedy I didn't know what your thoughts were, but I started by just [enumerating null](https://github.com/OAI/OpenAPI-Specification/issues/1900#issuecomment-578409376). Open to any and all other ideas!

### DIFF
--- a/job_spec/WATER_MAP.yml
+++ b/job_spec/WATER_MAP.yml
@@ -73,6 +73,7 @@ WATER_MAP:
           - iterative
           - logstat
           - nmad
+          - null
           - numpy
           - None
     water_level_sigma:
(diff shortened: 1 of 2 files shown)


jhkennedy (approved):
@jacquelynsmale yes, this is precisely the change I was going to circle back to make and was specifically what I was worried about. Thanks!

PR #1645, opened 2023-05-25
Add `INSAR_ISCE_BURST` job

jtherrmann:
TODO:

- [x] Consider these as third-party granules [here](https://github.com/ASFHyP3/hyp3/blob/develop/apps/api/src/hyp3_api/validation.py#L68) until they are present in CMR.
- [x] Add Changelog entry.
- [x] Should `azimuth_looks` and `range_looks` be restricted to certain values?
- [x] Pick a `timeout` value

### DIFF
--- a/apps/api/src/hyp3_api/validation.py
+++ b/apps/api/src/hyp3_api/validation.py
@@ -66,6 +66,7 @@ def get_cmr_metadata(granules):
 
 
 def is_third_party_granule(granule):
+    # FIXME remove the -BURST case when we integrate with prod CMR
     return granule.startswith('S2') or granule.startswith('L') or granule.endswith('-BURST')
 
 
--- a/apps/api/src/hyp3_api/validation.py
+++ b/apps/api/src/hyp3_api/validation.py
@@ -66,7 +66,7 @@ def get_cmr_metadata(granules):
 
 
 def is_third_party_granule(granule):
-    return granule.startswith('S2') or granule.startswith('L')
+    return granule.startswith('S2') or granule.startswith('L') or granule.endswith('-BURST')
 
 
 def check_granules_exist(granules, granule_metadata):
--- a/job_spec/INSAR_ISCE_BURST.yml
+++ b/job_spec/INSAR_ISCE_BURST.yml
@@ -1,0 +1,56 @@
+INSAR_ISCE_BURST:
+  required_parameters:
+    - granules
+  parameters:
+    granules:
+      default:  '""'
+      api_schema:
+        type: array
+        minItems: 2
+        maxItems: 2
+        example:
+          - TODO
+          - TODO
+        items:
+          description: TODO
+          type: string
+          pattern: "^S1_.*-BURST"
+          minLength: 43
+          maxLength: 43
+          example: TODO
+    bucket_prefix:
+      default:  '""'
+    azimuth_looks:
+      description: Number of looks to take in azimuth
+      type: integer
+      default: 4
+      # TODO enum?
+    range_looks:
+      description: Number of looks to take in range
+      type: integer
+      default: 20
+      # TODO enum?
+  validators: []
+  tasks:
+    - name: ''
+      image: ghcr.io/asfhyp3/hyp3-isce2
+      command:
+        - ++process
+        - insar_tops_burst
+        - ++omp-num-threads
+        - '1'
+        - --bucket
+        - '!Ref Bucket'
+        - --bucket-prefix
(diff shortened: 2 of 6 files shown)


forrestfwilliams:
This looks good to me! To answer your outstanding questions:

1. `azimuth_looks` should be 4, 2, or 1 and `range_looks` should be 20, 10, 5.
2. I think a timeout of 10 minutes would be appropriate for now. We can adjust it if the seems too lax or tight.

job_spec/INSAR_ISCE_BURST.yml
```diff
@@ -0,0 +1,56 @@
...
+        - --bucket
+        - '!Ref Bucket'
+        - --bucket-prefix
+        - Ref::bucket_prefix
+        - --azimuth-looks
+        - Ref::azimuth_looks
+        - --range-looks
+        - Ref::range_looks
+        - Ref::granules
+      timeout: 3600  # TODO decide on appropriate value
```
forrestfwilliams, line 51:
```suggestion
      timeout: 600
```

job_spec/INSAR_ISCE_BURST.yml
```diff
@@ -0,0 +1,56 @@
...
+          minLength: 43
+          maxLength: 43
+          example: S1_249434_IW1_20230523T170733_VV_8850-BURST
+    bucket_prefix:
+      default:  '""'
+    azimuth_looks:
+      description: Number of looks to take in azimuth
+      type: integer
+      default: 4
+      # TODO enum?
```
forrestfwilliams, line 27:
```suggestion
      enum:
          - 4
          - 2
          - 1
```

job_spec/INSAR_ISCE_BURST.yml
```diff
@@ -0,0 +1,56 @@
...
+    range_looks:
+      description: Number of looks to take in range
+      type: integer
+      default: 20
+      # TODO enum?
```
forrestfwilliams, line 32:
```suggestion
      enum:
          - 20
          - 20
          - 5
```

apps/api/src/hyp3_api/validation.py
```diff
@@ -66,7 +66,7 @@ def get_cmr_metadata(granules):
 
 
 def is_third_party_granule(granule):
-    return granule.startswith('S2') or granule.startswith('L')
+    return granule.startswith('S2') or granule.startswith('L') or granule.endswith('-BURST')
```
asjohnston-asf, line 69: I think this might deserve a #FIXME, or some other way to call out that it's temporary until we integrate hyp3-isce2 with production CMR

job_spec/INSAR_ISCE_BURST.yml
```diff
@@ -0,0 +1,53 @@
...
+        type: array
+        minItems: 2
+        maxItems: 2
+        example:
+          - S1_249434_IW1_20230523T170733_VV_8850-BURST
+          - S1_249434_IW1_20230511T170732_VV_07DE-BURST
+        items:
+          description: The reference and secondary bursts.
+          type: string
+          pattern: "^S1_.*-BURST"
```
asjohnston-asf, line 17: @forrestfwilliams I think we should only support `IW` burst granules, and exclude `EW`?

job_spec/INSAR_ISCE_BURST.yml
```diff
@@ -0,0 +1,53 @@
...
+      default:  '""'
+      api_schema:
+        type: array
+        minItems: 2
+        maxItems: 2
+        example:
+          - S1_249434_IW1_20230523T170733_VV_8850-BURST
+          - S1_249434_IW1_20230511T170732_VV_07DE-BURST
+        items:
+          description: The reference and secondary bursts.
```
asjohnston-asf, line 15: I think this is technically the description for an individual item, so "The name of the Sentinel-1 SLC IW burst granule to process" would be slightly more accurate

apps/api/src/hyp3_api/validation.py
jtherrmann: Added a `FIXME` and an issue: https://github.com/ASFHyP3/hyp3/issues/1651

job_spec/INSAR_ISCE_BURST.yml
jtherrmann: This is included in the future goals listed in the JIRA ticket.

asjohnston-asf (approved):

asjohnston-asf (approved):

Issue #1666, opened 2023-06-15
Subscription worker lambda fails with `Runtime.ImportModuleError`


asjohnston-asf:
`Runtime.ImportModuleError: Unable to import module 'subscription_worker': cannot import name 'DEFAULT_CIPHERS' from 'urllib3.util.ssl_' (/var/task/urllib3/util/ssl_.py)`

Likely introduced with `requests v2.30.0` released May 4. We saw a similar error for the HyP3 API lambda when upgrading from requests==2.29.0 to 2.31.0 : https://github.com/ASFHyP3/hyp3/pull/1654 . More details about the conflict between requests and botocore at https://github.com/psf/requests/issues/6438

asjohnston-asf:
Here's count of production subscription worker lambda failures per 5 minutes for the last three months, suggesting the issue was introduced in production with https://github.com/ASFHyP3/hyp3/releases/tag/v3.5.0 on May 5, the first production release following the requests 2.30.0 release on May 4.
![Screenshot from 2023-06-14 16-52-39](https://github.com/ASFHyP3/hyp3/assets/17994518/6c38ffe4-7d2c-4e79-b732-0b6cfbba55fb)

asjohnston-asf:
This error is due to incompatibility between `requests>=2.30.0` and `botocore<1.29.123`. Troubleshooting revealed an invalid `install_requires` block in `lib/dynamo/setup.py`. This resulted in the latest-and-greatest boto3/botocore not being included in the subscription worker lambda deployment package, so the deployed code was pulling in the default `botocore=1.29.90` in the [Python 3.9](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html) lamba runtime.

PR #1667, opened 2023-06-15
fix invalid install_requires clause in dynamo/setup.py

asjohnston-asf:
Fixes https://github.com/ASFHyP3/hyp3/issues/1666


asjohnston-asf:
This likely also allows https://github.com/ASFHyP3/hyp3/pull/1655 to be merged without introducing the same import error in the HyP3 API lambda.

hjkristenson (approved):

PR #1714, opened 2023-07-17
Fix WKT validator

jtherrmann:
TODO:

- [ ] We fixed `test_get_subscription_by_id` by adding missing subscription fields. Why is this not an issue on `develop`?
- [ ] We may be able to fix some other tests the same way, see https://github.com/ASFHyP3/hyp3/pull/1490
- [ ] Fix remaining tests


jtherrmann:
We no longer require WKT validation since removing the Subscriptions feature.

PR #1725, opened 2023-07-20
Fargate

asjohnston-asf:
The architecture changes are in apps/compute-cf.yml.j2 , plus updates to the AWS::Batch::JobDefinition in apps/workflow-cf.yml.j2 . The other changes are removing the now-obsolete AMI ID and Instance Type stack parameters, and updating the memory/vcpu/disk requirements for each job type.

Fargate compute environments do not have a `desiredvCpus` attribute like EC2 compute environments, mandating changes to scale_cluster.py. Fargate environments appear to "do the right thing" when lowering the `maxVcpus` parameter, eliminating the need to disable the environment until `desiredVcpus <= new maxVcpus`.

The allowed CPU/Memory combinations for AWS Fargate tasks are documented at https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-resourcerequirement.html

I've picked the memory value that's closest to the existing value for each job spec definition, then picked the fewest CPUs available for that much memory. Much of the work will be evaluating whether the slightly-modified memory settings are still a good fit for each job type, particularly for job types where available memory decreased slightly.

Available disk ranges from 21 GB to 200 GB. I've not yet put much effort into pinpointing the disk requirement for each job type. Right-sizing memory will have a bigger cost/performance impact than right-sizing disk, so just making sure each job type has enough disk to succeed at all is likely sufficient for an initial release.

# TODO

- [x] remove `fargate` from list of deployment branches in .github/workflows/deploy-enterprise-test.yml
- [x] remove `fargate` from list of branch protection rules for `hyp3-enterprise-test` environment in GitHub
- [ ] request Fargate Spot vcpu quota increase for hyp3-its-live account


apps/workflow-cf.yml.j2
```diff
@@ -55,11 +57,15 @@ Resources:
           {% endif %}
         JobRoleArn: !Ref TaskRoleArn
         ExecutionRoleArn: !GetAtt ExecutionRole.Arn
+        NetworkConfiguration:
+          AssignPublicIp: ENABLED
```
asjohnston-asf, line 61: A public IP is required for tasks to have internet access in our typical deployments using the default VPC. It may not be necessary in Earthdata Cloud environments using a private VPC.

```suggestion
        {% if security_environment != 'EDC' %}
        NetworkConfiguration:
          AssignPublicIp: ENABLED
        {% endif %}
```

asjohnston-asf: Did an ad-hoc deployment into the edc uat account and things worked fine. The help text when creating a job definition suggests this setting is required for outbound internet access, regardless of the VPC's network configuration:

> Screenshot from 2023-07-22 15-17-45

apps/scale-cluster/src/scale_cluster.py
jtherrmann, line 1: I'm curious to chat about why the scaling method has changed.

apps/step-function.json.j2
jtherrmann, line 1: Remind me why the memory values have decreased?

apps/workflow-cf.yml.j2
jtherrmann, line 61: What's this for?

apps/step-function.json.j2
asjohnston-asf: Fargate jobs are restricted to specific cpu/memory combinations: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-resourcerequirement.html

apps/workflow-cf.yml.j2
asjohnston-asf: Fargate jobs running in a public subnet require a public ip address for outbound internet access, see https://docs.aws.amazon.com/AmazonECS/latest/userguide/fargate-task-networking.html

I'm not sure if fargate jobs running in a private subnet with a NAT (e.g. Earthdata Cloud) require a public ip, but I did an ad-hoc deployment into our earthdata cloud uat account and jobs ran fine there with this setting enabled.

jtherrmann (approved):

PR #1752, opened 2023-08-07
Remove c5 instances with less disk for INSAR ISCE jobs

cmarshak:
Based on discussion here: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/issues/145#issuecomment-1668251751

Need extra disk space based on new generatin of instsances. Note the access instances are restricted to the new instance type with 118 GB of space.


asjohnston-asf (approved):

asjohnston-asf:
Not currently, no. HyP3 uses "d" instance types that include Instance Storage in proportion to their vCPUs/memory. This is the first workflow I recall where disk space be the limiting factor. Memory has typically been the bottleneck, so we haven't invested in letting jobs allocate a custom amount of disk space. It's doable, but we'd have to set aside time to develop the feature.

PR #1776, opened 2023-08-17
Remove the Subscriptions feature

jtherrmann:
## Overview

I confirmed that each of the following commands shows no output:

```bash
# finds all tracked filenames containing "subscription" (case-insensitive)
git ls-tree -r remove-subscriptions --name-only | grep -i subscription

# finds all occurrences of "subscription" (case-insensitive) in tracked files other than the Changelog
git grep -i subscription | grep -v CHANGELOG
```

We expect these resources to get deleted:
- subscription manager and worker lambdas
- subscriptions dynamodb table
- subscription_id parameter for get jobs api endpoint
- subscriptions api endpoints
- subscriptions-related api schemas

## Before merging to test

- [x] Changelog entry.

Verify changes in a sandbox deployment:

- [x] Create a new sandbox deployment (branch off of `develop`).
- [x] Set user quota for myself.
- [x] Create some subscriptions in the sandbox and let them generate some products.
- [x] Merge `remove-subscriptions` into `subscriptions-removal-sandbox` and verify that all expected resources get deleted
- [x] Why does Swagger UI show INVALID for API schema? (see bottom-right corner of page)
- [x] Manually remove the `subscription_id` field from the Jobs table.

## After merging to test

- [ ] Verify that all expected resources get deleted
- [ ] Verify that Swagger UI shows VALID for API schema (see bottom-right corner of page)
- [ ] Manually remove the `subscription_id` field from the Jobs table.


apps/api/src/hyp3_api/api-spec/job_parameters.yml.j2
asjohnston-asf, line 1: We can remove the `job_specification` schema. I suspect all references have already been removed but haven't double-checked yet.

apps/api/src/hyp3_api/routes.py
asjohnston-asf, line 1: I suspect the `WKTValidator` class is no longer needed.

apps/api/src/hyp3_api/api-spec/job_parameters.yml.j2
jtherrmann: Good catch, I removed the schema and confirmed that all references have been removed.

apps/api/src/hyp3_api/routes.py
jtherrmann: Good catch, I removed it. I should probably find a way to do a comprehensive sweep for any unused functions and classes.

jtherrmann: I ran https://pypi.org/project/vulture/ on `apps/` and `lib/`, on both `develop` and `remove-subscriptions`, to confirm that no more unused code has appeared on the `remove-subscriptions` branch.

asjohnston-asf (approved):

PR #1784, opened 2023-08-22
update INSAR_ISCE job type for DockerizedTopsApp v0.2.4 release

asjohnston-asf:
Updates INSAR_ISCE job spec to match INSAR_ISCE_TEST to reflect the upcomming [DockerizedTopsApp](https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp) v0.2.4 release.

The only differences between INSAR_ISCE and INSAR_ISCE_TEST will be the job type name and the docker image tag used to run jobs:
```
$ diff INSAR_ISCE.yml INSAR_ISCE_TEST.yml
1c1
< INSAR_ISCE:
---
> INSAR_ISCE_TEST:
69a70
>       image_tag: test
96a98
>       image_tag: test
```

### DIFF
--- a/.github/workflows/deploy-enterprise.yml
+++ b/.github/workflows/deploy-enterprise.yml
@@ -39,7 +39,7 @@ jobs:
             product_lifetime_in_days: 180
             quota: 0
             job_files: job_spec/INSAR_ISCE.yml job_spec/INSAR_ISCE_TEST.yml
-            instance_types: c6id.xlarge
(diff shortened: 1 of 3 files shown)


jhkennedy (approved):
We'll want to double check all the secrets are available for these deployments (I believe they are already)

asjohnston-asf:
Confirmed that the secret values for raider are available in all three deployments

Issue #1765, opened 2023-08-10
Specify JWT auth public key as GitHub secret


jtherrmann:
**Jira:** https://asfdaac.atlassian.net/browse/TOOL-2127

*Note: The above link is accessible only to members of ASF.*

----------------------------------------------------------------------------------------------------

Last time that we had to update the public key as part of https://github.com/ASFHyP3/hyp3/pull/1763, we had to manually update the CloudFormation parameter via the AWS console for every HyP3 deployment, because changing the default parameter value in the template does not update the value in the stack. We should specify the value as a GitHub secret so that we can update it automatically in the future.

PR #1802, opened 2023-08-28
Specify jwt auth public key as github secret

jtherrmann:
See https://github.com/ASFHyP3/hyp3/issues/1765

Before merging:

- [x] Add `AUTH_PUBLIC_KEY` as a repository secret (use the value from https://sentinel1.asf.alaska.edu/pubkey)

After merging:

- [ ] Authenticate to the `hyp3-test` API with a fresh cookie and confirm I can query the endpoints


AndrewPlayer3 (approved):
Looks good!

PR #1814, opened 2023-09-01
Reduce product lifetime to 14 days to reduce storage spending while idle

jhkennedy:


.github/workflows/deploy-enterprise.yml
```diff
@@ -51,7 +51,7 @@ jobs:
             domain: hyp3-tibet-jpl.asf.alaska.edu
             template_bucket: cf-templates-1or0efwqffkgd-us-west-2
             image_tag: latest
-            product_lifetime_in_days: 60
+            product_lifetime_in_days: 14
```
jhkennedy, line 54:
```suggestion
            product_lifetime_in_days: 7
```

Maybe 7 is more appropriate as charlie said "in the next week or so"

jhkennedy:
Nevermind -- leaving as is for now.

PR #1708, opened 2023-07-13
Upgrade openapi

jtherrmann:
Branched from https://github.com/ASFHyP3/hyp3/pull/1380, uses latest package versions.

See https://github.com/ASFHyP3/hyp3/issues/1193

Issue #1193, opened 2022-09-09
Upgrade openapi packages


asjohnston-asf:
**Jira:** https://asfdaac.atlassian.net/browse/TOOL-2046

After upgrading to openapi-spec-validator to v0.5.0, `make install` fails with:

```
The conflict is caused by:
    The user requested openapi-spec-validator==0.5.1
    openapi-core 0.14.5 depends on openapi-spec-validator<0.5.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict


Pip subprocess error:
ERROR: Cannot install -r /home/asjohnston/src/hyp3/requirements-apps-api.txt (line 4) and openapi-spec-validator==0.5.1 because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts

failed

CondaEnvException: Pip failed
```

`requirements-api.txt` requires `openapi-core 0.14.5`, which depends on openapi-spec-validator<0.5.0. See https://github.com/p1c2u/openapi-core/issues/402 for more details.

We separately specify openapi-spec-validator as a requirement in `requirements-all.txt`, since we use it for static analysis independent of deploying the API. We can't upgrade the version specified there past 0.4.0 because of this conflict.

asjohnston-asf:
The conflict between openapi-spec-validator and openapi-core has been resolved in the current versions, but upgrading both leads to another dependency conflict for `jsonschema`:

```
The conflict is caused by:
    cfn-lint 0.72.5 depends on jsonschema<5 and >=3.0
    openapi-spec-validator 0.5.0 depends on jsonschema<5.0.0 and >=4.0.0
    aws-sam-translator 1.55.0 depends on jsonschema~=3.2
```

This should be resolved with the next aws-sam-translator release. https://github.com/aws/serverless-application-model has already relaxed the jsonschema requirement in `develop`, so this conflict should be resolves with the next aws-sam-translator release.

https://github.com/aws/serverless-application-model/issues/2426

jtherrmann:
I built my environment from scratch on branch `dependabot/pip/openapi-core-0.16.4` (see https://github.com/ASFHyP3/hyp3/pull/1380). I confirmed the following versions of the `openapi-*` packages:

```
openapi-core              0.16.4
openapi-schema-validator  0.3.4
[...]

PR #1821, opened 2023-09-06
Upgrade openapi after subscriptions removal

jtherrmann:
See https://github.com/ASFHyP3/hyp3/issues/1193

After removing the Subscriptions feature, we've significantly changed our API spec and have removed the need for WKT string validation, which makes much of the work done on https://github.com/ASFHyP3/hyp3/pull/1708 obsolete, so I'm starting over from scratch with this PR.

This PR includes the following changes:

- Upgrade the `openapi-core`, `openapi-spec-validator`, and `jsonschema` packages to their latest versions
- Minor updates to the reference paths in our OpenAPI spec files
- Convert our API endpoints from Flask class-based views to functions decorated with `@app.route`, which leads to simpler code (in my opinion) and, more importantly, allows us to more easily use the `FlaskOpenAPIViewDecorator` decorator provided by `openapi-core`, which in turn allows us to re-implement the following features according to the latest `openapi-core` interface:
    - Disabling of response validation
    - Custom error formatting

TODO:
- [ ] Create an issue for converting the `openapi-core` pin back to the normal format following the next PyPI release (see the `TODO` comment in the requirements file)


jtherrmann:
The `cfn-lint` package pins `jsonschema` [here](https://github.com/aws-cloudformation/cfn-lint/blob/main/setup.py) (see https://github.com/aws-cloudformation/cfn-lint/pull/2792).

jtherrmann:
The version conflict should be resolved by the most recent version of `cfn-lint`.

jtherrmann:
We can now upgrade the `openapi-*` and `jsonschema` packages to their latest versions.

apps/api/src/hyp3_api/api-spec/job_parameters.yml.j2
```diff
@@ -53,7 +53,7 @@ components:
           enum:
             - {{ job_type }}
         name:
-          $ref: "./openapi-spec.yml#components/schemas/name"
+          $ref: "./openapi-spec.yml#/components/schemas/name"
```
forrestfwilliams, line 56: I see the additional of this `/` in several files. Is this an update to how our yamls work?

apps/api/src/hyp3_api/routes.py
```diff
@@ -68,7 +67,7 @@ def render_ui():
 
 
 @app.errorhandler(404)
-def error404(e):
+def error404(_):
```
forrestfwilliams, line 70: I'm not a fan of an `_` as an input argument name. Is this standard practice? Could we use something like `unused_arg` instead?

forrestfwilliams (commented):
From what I can tell, this PR look good. I'm confident in the changes to our configuration YAMLs, the dependabot changes, the changelog, and requirements files.

However, I have not worked with Flask APIs in the past, and thus don't feel good rubber stamping the changes in `apps/api/src/hyp3_api/routes.py`. Would @jhkennedy or @asjohnston-asf  be willing to lend a hand here?

apps/api/src/hyp3_api/api-spec/job_parameters.yml.j2
jtherrmann: I believe it's just an update to how `jsonschema` expects reference paths to be written. Just a syntax update, nothing more.

apps/api/src/hyp3_api/routes.py
jtherrmann: I believe it is standard practice, and it gets rid of PyCharm's warning about an unused parameter.

apps/api/src/hyp3_api/routes.py
```diff
@@ -93,67 +92,57 @@ def default(self, o):
...
+    errors_handler_cls=ErrorHandler,
+)
+
 
-    def post(self):
-        return jsonify(handlers.post_jobs(request.get_json(), g.user))
+@app.route('/jobs', methods=['POST'])
+@openapi
+def jobs_post():
+    return jsonify(handlers.post_jobs(request.get_json(), g.user))
```
jhkennedy, line 117: This is completely nitpicky but I like the methods in alphabetical order for a route

jhkennedy (approved):
Looks great to me! I like the class to decorator change.

apps/api/src/hyp3_api/routes.py
jtherrmann: Fair, I'm just going to merge since I'm eager to get this deployed, but will keep in mind for the future.

PR #1859, opened 2023-10-02
Fill COP30 Coverage Gaps

AndrewPlayer3:
This PR adds a new geojson file which allows for jobs to be ran over Armenia and Azerbaijan.


asjohnston-asf:
I think we can overwrite the existing `dem_coverage_map_cop30.geojson`. Since it's bundled with the deployed application we'll be able to deploy/validate the changes to test without impacting prod.

It's also worth simplifying the new geojson by taking the union (i.e. [shapely.unary_untion()](https://shapely.readthedocs.io/en/stable/reference/shapely.unary_union.html)) of all the individual tiles. Should reduce the size of the file from ~3 MB to ~1 MB and make the intersection check faster.

asjohnston-asf:
Looks great, will be ready to merge once https://github.com/ASFHyP3/hyp3/pull/1862 is merged to develop and synced with this branch so that the tests pass.

asjohnston-asf (dismissed):

asjohnston-asf (approved):

PR #1876, opened 2023-10-12
Add `phase_filter_parameter` for `INSAR_GAMMA`

jtherrmann:


job_spec/INSAR_GAMMA.yml
```diff
@@ -63,6 +63,14 @@ INSAR_GAMMA:
         enum:
           - 20x4
           - 10x2
+    phase_filter_parameter:
+      api_schema:
+        description: Adaptive phase filter exponent (alpha). Larger values result in more aggressive filtering. If zero, adaptive phase filter will be skipped. We recommend against values less than 0.2, or values with more than two decimal places of precision.
+        type: number
+        format: float
```
jacquelynsmale, line 70: Is there a reason you want to specify float format?

jacquelynsmale (approved):

job_spec/INSAR_GAMMA.yml
jtherrmann: I noticed this option for the first time when I was reading the docs: https://swagger.io/docs/specification/data-models/data-types/#numbers

> An optional `format` keyword serves as a hint for the tools to use a specific numeric type

I'm not sure what it does but I'm interested to see if it shows up in our Swagger interface.

jtherrmann: It shows up like this in Swagger:

```
number($float)
default: 0.6
maximum: 1
minimum: 0
```

Not super exciting. I might have expected it to display min and max as floats.

jtherrmann: Ah, this places restrictions on what the user can enter, which is bad. See https://github.com/ASFHyP3/hyp3/pull/1877

PR #1882, opened 2023-10-20
Improve `phase_filter_parameter` description

jtherrmann:


hjkristenson (approved):
I think I actually like `more aggressive filtering` better than `stronger filtering`. Maybe we should change out the phrasing in the other documentation... I'm okay with it either way, though.

PR #1900, opened 2023-10-26
lower job memory reservations to accomodate AL2023 AMI

asjohnston-asf:
The ECS Agent running on the AL2023 AMI reserves slightly more memory than on the AL2 AMI. This leaves only 31,517 MB available for jobs on an `r6id.xlarge`; 126,722 available on an `r6id.4xlarge`, ect.

This PR lowers the memory reserved for jobs to 7,875 MB/vcpu so they'll fit on any of the `r6id` family.

Screenshot from 2023-10-25 17-16-57

# TODO
- [ ] double-check 10m rtc memory reservation

### DIFF
--- a/apps/step-function.json.j2
+++ b/apps/step-function.json.j2
@@ -138,7 +138,7 @@
         "ResourceRequirements": [
           {
             "Type": "MEMORY",
-            "Value": "7900"
+            "Value": "7800"
           }
         ]
       },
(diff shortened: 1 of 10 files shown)


asjohnston-asf:
FYI, for comparison, I confirmed 31,603 MB of memory is available to jobs when running either the AL2 AMI or the NGAP AMI on an `r6idn.xlarge`.

jhkennedy (approved):

jhkennedy:
We should run the autoRIFT golden test to ensure optical jobs still have enough memory. And maybe check the GUNWs as well (we should add a golden test for them)

PR #1922, opened 2023-11-10
Provide ESA creds to insar and rtc jobs

jtherrmann:
This will allow testing the new `hyp3lib` orbit fetching implementation on `hyp3-test`. Afterwards, we can either revert this PR (if we're not going to release the new orbit fetching implementation), or add a Changelog entry and release these changes (adding the secrets for any other job types that need them, as well).


jtherrmann:
Looks like we already added `ESA_USERNAME` and `ESA_PASSWORD` to Secrets Manager in UAT at some point, I assume in preparation for this work.

jacquelynsmale (approved):

PR #1926, opened 2023-11-13
Auto-Correct Private DNS Names Networking Issue

forrestfwilliams:
This is a first attempt at a solution for the Private DNS names error that affects HyP3 ISCE2 burst jobs. The strategy is:

- Add a cron lambda to HyP3 that checks the status of the offending parameter
- Inside the lambda:
    1. If the parameter is set to `True`, change it to `False`
    2. If it is already `False`, do nothing

PR #1939, opened 2023-11-20
Revert "Auto-Correct Private DNS Names Networking Issue"

jtherrmann:
Reverts ASFHyP3/hyp3#1926

TODO:

- [x] Wait until the permanent fix is released.
- [x] Don't delete the Changelog entry, and add a new entry for the revert.


CHANGELOG.md
jtherrmann, line 1: We should not remove the Changelog entry, and we should add another entry for the revert.

forrestfwilliams:
Will revert this once the permanent fix is available from NGAP.

jhkennedy (approved):

PR #1981, opened 2023-12-20
Release v4.5.1 -- output_resolution numeric type fix

jhkennedy:
Test `INSAR_ISCE_TEST` jobs:
* :x: with no `output_resolution` specified:
  https://hyp3-enterprise-test.asf.alaska.edu/jobs?user_id=jhkennedy&name=Los-Angeles-0_64_HRRR_1121_part-deux
  * these failed due to the job timeout being hit, but did start running succfully, which they did not in v4.5.0, so really, :heavy_check_mark: for this PR
* :heavy_check_mark:  with `output_resolution` specified as an int:
  https://hyp3-enterprise-test.asf.alaska.edu/jobs/2b718b83-7684-449b-9db2-a6c9d6dc7e41
* :heavy_check_mark: and I get the expected rejection with a float `output_resolution`:
  image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [4.5.1]
+### Fixed
+- `output_resolution` in the `INSAR_ISCE_TEST` job spec is now correctly specified as an int instead of number, which can be a float or an int.
+
 ## [4.5.0]
 ### Changed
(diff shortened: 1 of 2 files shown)


jhkennedy:
The no `output_resolution` specified batch of jobs all failed by hitting the timeout, getting 17%-ish through in 6 hours. 

Notably, we are running on `c6id` instances in the JPL deployment and have `++omp-num-threads 4` set in the job spec, but in enterprise-test we are running on `r6id` instances which have 1/2 the CPUs available for the same memory reservation.

I suspect the timeout was caused either all or in part by CPU contention with multiple jobs running, but it could also  just be due to new options/layers enabled.

I think we should just ship this so we can test the new options @ JPL *and we really should set up test deployments for our science envs.*

jtherrmann (approved):

PR #2007, opened 2024-01-11
Openapi validate server

jtherrmann:


jtherrmann:
The solution is actually much simpler than this. Follow-up PR incoming.

Issue #2003, opened 2024-01-10
Bump hyp3's openapi-core version pin after the next openapi-core release


jtherrmann:
After the next [openapi-core release](https://github.com/python-openapi/openapi-core/releases) beyond the current latest pre-release of v0.19.0a1, we should update https://github.com/ASFHyP3/hyp3/blob/develop/requirements-apps-api.txt to pin `openapi-core` to the latest release (I'm not sure if dependabot will bump it correctly since it's currently pinned to the pre-release on GitHub, since pre-releases are not available on PyPI).

jtherrmann:
With the release of https://github.com/python-openapi/openapi-core/releases/tag/0.19.0, this is now actionable and should probably be done sooner rather than later.

PR #2010, opened 2024-01-12
Release v5.0.1

jtherrmann:
See https://github.com/ASFHyP3/hyp3/issues/1193 for context

TODO:

- [x] Verify that `/user` works for both <https://hyp3-test-api.asf.alaska.edu> and <https://hyp3-enterprise-test.asf.alaska.edu> to verify the changes from https://github.com/ASFHyP3/hyp3/pull/2009
- [x] Verify that all API endpoints still work as expected (for good and bad requests) via Swagger and the HyP3 SDK
- [x] Verify that the API schemas still render correctly in Swagger.
- [ ] After release, verify that EDC prod API works correctly for both Swagger and SDK, and the same for the JPL deployments (in case there's any weird infrastructure in the JPL accounts that would modify our request URLs similar to the CloudFront issue in EDC)


requirements-apps-api.txt
```diff
@@ -1,7 +1,8 @@
 flask==2.2.5
 Flask-Cors==4.0.0
-jsonschema==4.17.3
-openapi-core==0.14.5
+jsonschema==4.20.0
+# TODO convert this pin back to the normal format after the next PyPI release
+openapi-core @ git+https://github.com/python-openapi/openapi-core@0.19.0a1
```
jhkennedy, line 5: Do we have an issue for this, or will dependabot pick it up?

jhkennedy (approved):
Looks good to me assuming the SwaggerUI and HyP3 SDK have been taken for a spin in both enterprise and UAT

requirements-apps-api.txt
jtherrmann: Yeah I created an issue for it because I'm not sure what dependabot will do: https://github.com/ASFHyP3/hyp3/issues/2003

PR #2036, opened 2024-01-24
Monthly credits reset

jtherrmann:
In this PR:
* Implement the monthly credits reset feature.
* Revise the `dynamo.user` module.

TODO:

- [x] Add/update unit tests when we're satisfied with the implementation.
- [x] Pass the `RESET_CREDITS_MONTHLY` env var through


.github/actions/deploy-hyp3/action.yml
```diff
@@ -38,8 +38,10 @@ inputs:
   DEFAULT_CREDITS_PER_USER:
     description: "The default number of credits given to a new user"
     required: true
+  RESET_CREDITS_MONTHLY:
```
asjohnston-asf, line 41: `RESET_CREDITS_EACH_MONTH` might be a little more expressive

apps/main-cf.yml.j2
```diff
@@ -36,6 +36,10 @@ Parameters:
     Type: Number
     MinValue: 0
 
+  ResetCreditsMonthly:
+    Description: "Reset each user's remaining credits each month if and only if this value is 'yes'."
+    Type: String
+
```
asjohnston-asf, line 42:
```suggestion
    AllowedValues:
      - yes
      - no

```

.github/actions/deploy-hyp3/action.yml
asjohnston-asf: It occurs to me that resetting could be driven by `DEFAULT_CREDITS_PER_USER != 0`. Thoughts on combining them into a single parameter? I kind of like the separate explicit parameter

lib/dynamo/dynamo/user.py
```diff
@@ -1,45 +1,93 @@
...
+
+
+def _reset_credits_if_needed(user: dict, default_credits: Decimal, current_month: str, users_table) -> dict:
+    if (
+            os.environ['RESET_CREDITS_MONTHLY'] == 'yes'
+            and user['month_of_last_credits_reset'] < current_month  # noqa: W503
+            and user['remaining_credits'] is not None  # noqa: W503
+    ):
+        user['month_of_last_credits_reset'] = current_month
+        user['remaining_credits'] = default_credits
```
asjohnston-asf, line 61: We're missing the ability to give individual users a different monthly credit allotment (previously implemented via user['max_jobs_per_month']
```suggestion
        user['remaining_credits'] = user.get('credits_per_month', default_credits)
```

tests/test_dynamo/test_user.py
```diff
@@ -1,73 +1,253 @@
...
-        'remaining_credits': 25,
+    user = dynamo.user._reset_credits_if_needed(
+        user=original_user_record,
+        default_credits=Decimal(25),
+        current_month='2024-02',
+        users_table=tables.users_table,
+    )
+
+    assert user == {'user_id': 'foo', 'remaining_credits': Decimal(10), 'month_of_last_credits_reset': '2024-01'}
+    assert tables.users_table.scan()['Items'] == [user]
```
asjohnston-asf, line 114: I tend to prefer using literal values as much as possible in tests, but I think re-using `original_user_record ` makes the asserts more expressive
```suggestion
    assert user == original_user_record
    assert tables.users_table.scan()['Items'] == [original_user_record]
```

tests/test_dynamo/test_user.py
```diff
@@ -1,73 +1,253 @@
...
+        default_credits=Decimal(25),
+        current_month='2024-02',
+        users_table=tables.users_table,
+    )
+
+    assert user == {'user_id': 'foo', 'remaining_credits': None, 'month_of_last_credits_reset': '2024-01'}
+    assert tables.users_table.scan()['Items'] == [user]
+
+
+def test_reset_credits_failed_same_month(tables, monkeypatch):
```
asjohnston-asf, line 155: `wrong_month` might be more accurate, since we're testing `2024-02` vs `2024-01`

asjohnston-asf (approved):

PR #2049, opened 2024-01-29
add INSAR_ISCE_BURST jobs to azdwr-hyp3

asjohnston-asf:


jhkennedy (approved):
Should we also add it to other deployments with INSAR jobs?

asjohnston-asf:
I'm inclined to say no, not unless asked.

PR #2058, opened 2024-02-01
fix cloudwatch metric used to monitor for api 5xx errors

asjohnston-asf:


apps/api/api-cf.yml.j2
```diff
@@ -44,8 +44,8 @@ Outputs:
   Url:
     Value: {{ '!Sub "https://${RestApi}.execute-api.${AWS::Region}.amazonaws.com/${Stage}/ui"' if security_environment == 'EDC' else '!Sub "https://${CustomDomainName}/ui"' }}
 
-  ApiId:
-    Value: !Ref RestApi
+  ApiName:
+    Value: !GetAtt RestApi.RestApiId
```
jtherrmann, line 48: Just confirming that this should be `RestApiId`, not `RestApiName`?

jtherrmann (approved):
LGTM, as long as that field name is correct (see my comment above).

apps/api/api-cf.yml.j2
asjohnston-asf: The available return values are documented at https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-restapi.html#aws-resource-apigateway-restapi-return-values

We want the auto-generated `Imported on 2022-03-15T20:13:40Z` name. I'm not 100% confident that's `RestApiId`, but it's the obvious one.

asjohnston-asf: :thinking: `!Ref RestApi` and `!GetAtt RestApi.RestApi.Id` ended up returning the same value. Deployment succeeded but the alarm still won't work. Needs more research and a followup PR.

PR #2061, opened 2024-02-01
quota yes/no strings to keep yaml parser from changing them to true/false

asjohnston-asf:
Without quotes, the CloudFormation yaml parser interprets yes/no as boolean true/false. The value of the applied tag in the S3 console ended up being the strings "true" or "false" instead of the strings "YES" or "NO".

Screenshot from 2024-02-01 12-33-22
Screenshot from 2024-02-01 12-33-35

### DIFF
--- a/apps/main-cf.yml.j2
+++ b/apps/main-cf.yml.j2
@@ -219,7 +219,7 @@ Resources:
       {% if security_environment == 'EDC' %}
       Tags:
         - Key: DAR
-          Value: YES
+          Value: "YES"
       {% endif %}
 
   LogBucketPolicy:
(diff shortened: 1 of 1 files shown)


asjohnston-asf:
successfully tested this with a simple template:

```
AWSTemplateFormatVersion: 2010-09-09

Resources:

  Bucket:
    Type: AWS::S3::Bucket
    Properties:
      Tags:
        - Key: FOO
          Value: YES
        - Key: BAR
          Value: NO
```

without quotes:
Screenshot from 2024-02-01 12-43-09

with double quotes:
Screenshot from 2024-02-01 12-41-56

jtherrmann (approved):

PR #2072, opened 2024-02-10
Configurable job costs

jtherrmann:
TODO:

- [x] Populate the default values for job parameters before we do the cost lookup
- [x] Fix all our tests that assume a job cost of 1
- [x] Confirm that credits are represented as `Decimal` everywhere
- [x] Convert `float` to `Decimal` when loading `costs.json`
- [x] Add/update tests for fractional costs
- [x] Render `costs.yml`
- [x] Add `/costs` API endpoint to return cost tables for Vertex?
- [x] Resolve https://github.com/ASFHyP3/hyp3/issues/2084
- [x] Add cost definitions to all job specs
- [x] Add `cost_profile` parameter to deployments
- [x] TODO/FIXME items in code
- [x] Changelog entry
- [x] Run the implementation by Disco
- [x] Compare cost tables against https://hyp3-docs.asf.alaska.edu/using/credits/
- [x] Create GitHub+Jira issue for serializing `Decimal` in the `/costs` response
- [x] Deploy to `credits-sandbox`
- [x] Increase `default_credits_per_user` parameter to `10000` in `hyp3-edc-uat`?


jtherrmann:
Deployed to `credits-sandbox`, submitted one job of each type and confirmed that credit cost and priority are what I expect: https://hyp3-credits-sandbox.asf.alaska.edu/jobs?user_id=jtherrmann&start=2024-02-19T18%3A30%3A00%2B00%3A00&end=2024-02-19T19%3A00%3A00%2B00%3A00

.github/workflows/deploy-credits-sandbox.yml
```diff
@@ -21,6 +21,7 @@ jobs:
             product_lifetime_in_days: 14
             default_credits_per_user: 0
             reset_credits_monthly: true
+            cost_profile: EDC_UAT
```
asjohnston-asf, line 24: `EDC` should be sufficient here, since I expect we'll ultimately use the same label for both EDC environments

apps/api/src/hyp3_api/api-spec/openapi-spec.yml.j2
```diff
@@ -106,6 +117,11 @@ paths:
 components:
   schemas:
 
+    # TODO specify more information?
```
asjohnston-asf, line 120: I can take this TODO this afternoon

apps/api/src/hyp3_api/routes.py
```diff
@@ -111,6 +111,13 @@ def __call__(self, errors):
 )
 
 
+# TODO tests
+@app.route('/costs', methods=['GET'])
+@openapi
+def costs_get():
+    return jsonify(handlers.get_costs())
```
asjohnston-asf, line 118: I'm not convinced a handler is necessary here. We could make this look like the /openapi.json handler above. It would mean importing dynamo.jobs.COSTS here, though, and I'm open to arguments that routes.py shouldn't need to know anything about the dynamo package.

apps/render_cf.py
```diff
@@ -63,6 +75,8 @@ def main():
             task['name'] = job_type + '_' + task['name'] if task['name'] else job_type
 
     render_default_params_by_job_type(job_types)
+    if args.cost_profile != 'None':
```
jtherrmann, line 78: Rely on `DEFAULT` cost profile instead.

apps/api/src/hyp3_api/api-spec/openapi-spec.yml.j2
asjohnston-asf: :thinking: Hrmm, OpenAPI isn't good at describing schemas where we're using job types / parameter values as dictionary keys, e.g.:

```
foo:
  cost: 1.0
bar:
  cost: 1.0
another:
  cost_parameter: hello
  costs:
    a: 1.0
    b: 2.0
```

That makes lookups in the implementation easy, but makes describing what to expect via OpenAPI hard. If we use lists instead of dictionaries it makes describing the schema simple, but makes lookups at runtime more cumbersome:

```
- job_type: foo
  cost: 1.0
- job_type: bar
  cost: 1.0
- job_type: another
  cost_parameter: hello
  costs:
    - parameter_value: a
      cost: 1.0
    - parameter_value: b
      cost: 2.0
 ```

Punting this one and leaving it as-is for the moment.

jtherrmann: For posterity, we realized we could do this with Jinja, but not worth the effort.

PR #2076, opened 2024-02-12
Populate job dict with default param values

jtherrmann:
TODO:

- [x] Create `default_params_by_job_type.json` during template rendering
- [x] Changelog
- [x] Review current default params implementation


asjohnston-asf:
Having default values populated in the database potentially eliminates the need to define parameters for the AWS Batch Job definitions at https://github.com/ASFHyP3/hyp3/blob/develop/apps/workflow-cf.yml.j2#L46 , as well as the `isPresent` checks for RTC_GAMMA resolution at https://github.com/ASFHyP3/hyp3/blob/develop/apps/step-function.json.j2#L78 and https://github.com/ASFHyP3/hyp3/blob/develop/apps/step-function.json.j2#L111 .

It does no harm to leave those in place, and we'll want to deliver the new default behavior to production and let all existing jobs cycle through before removing them.

asjohnston-asf (approved):

PR #2081, opened 2024-02-13
use reusable-release-checklist-comment workflow

asjohnston-asf:


jtherrmann (approved):
Nice catch, I'm surprised we hadn't done this already.

PR #2166, opened 2024-03-25
Switch to prod and test ITS_LIVE deployments

jhkennedy:
Instead of using a test job type, this stands up a test stack.

I've gone ahead and setup the [`hyp3-its-live-test` environment](https://github.com/ASFHyP3/hyp3/settings/environments/2553247154/edit) and updated the SSL cert to cover hyp3-its-live and hyp3-its-live-test (dropping hyp3-its-live-eu). 

I've also reduced the product lifetime in the production hyp3-its-live deployment


docs/deployments/JPL-deployment-policy-cf.yml
```diff
@@ -30,6 +30,7 @@ Resources:
               - s3:*
               - secretsmanager:*
               - sns:*
+              - sqs:*
```
jhkennedy, line 33: This was needed for the `its-live-monitoring` stacks

.github/workflows/deploy-enterprise.yml
```diff
@@ -18,13 +18,12 @@ jobs:
             domain: hyp3-its-live.asf.alaska.edu
             template_bucket: cf-templates-3o5lnspmwmzg-us-west-2
             image_tag: latest
-            product_lifetime_in_days: 45
+            product_lifetime_in_days: 14
```
jhkennedy, line 21: How that we automagically publish products to the ITS_LIVE AWS Open Data bucket as the last step in AUTORIFT jobs, we don't need to retain products long term in this deployment

.github/workflows/deploy-enterprise-test.yml
```diff
@@ -31,7 +31,6 @@ jobs:
               job_spec/INSAR_ISCE_BURST.yml
               job_spec/WATER_MAP.yml
               job_spec/WATER_MAP_EQ.yml
-              job_spec/S1_CORRECTION_ITS_LIVE.yml
```
jhkennedy, line 34: this job isn't available outside of the ITS_LIVE deployments, so we don't need to test it here.

forrestfwilliams (approved):

PR #2173, opened 2024-03-28
Release v6.5.1

jtherrmann:
TODO:

- [x] Verify the lambda function is working as expected in UAT.


forrestfwilliams (approved):
All the code LGTM. I've checked and the lambda was deployed correctly in UAT. I also manually enabled Private DNS for the endpoint in question and verified that the lambda successfully disabled this setting.

PR #2192, opened 2024-04-09
add a raider job for the ARIA project

jhkennedy:
The ARIA project would like to be able to run the RAiDER step independently of the ISAR_ISCE job step in order to:
* update products with a different weather model (e.g., none -> HRRR or ERA5 -> HRRR)
* re-run just the RAiDER step for failed INSAR_ISCE jobs which successfully completed the DockerizedTopsApp step

TODO requirements for this PR:
- [x] update RAiDER's hyp3 entry point to take an alternate `--input-bucke-prefix` argument: 
   https://github.com/dbekaert/RAiDER/pull/636

TODO's post-release of this PR:
- [ ] Update the ingest workflow to also look at ARIA_RAIDER jobs


job_spec/ARIA_RAIDER.yml
```diff
@@ -0,0 +1,53 @@
...
+      image: ghcr.io/dbekaert/raider
+      command:
+        - ++process
+        - calcDelaysGUNW
+        - --bucket
+        - '!Ref Bucket'
+        - --bucket-prefix
+        - Ref::bucket_prefix
+        - --job-id
+        - Ref::job_id
```
jhkennedy, line 38: I believe we'll need to update RAiDER to take an additional parameter for this.

I had considered just passing the provided `job_id` to RAiDER as `--bucket-prefix`, and that *will* run through, but it would upload the competed products back to the previous INSAR_ISCE job-id prefix, not this job's prefix, in the hyp3 content bucket. So we need a way to get data from a different job-id prefix than the one we upload resulting products to.

jhkennedy: Thinking more about this, instead of taking a HyP3 Job ID; this could take a URL to a GUNW product, so it can work on HyP3 Jobs or Published products.

If we support running it on published products, we'll want RAiDER to update the version number and *not* clobber the already published GUWN in the archive

jhkennedy: Discussing it more with the ARIA team, we'll just focus on processing a previously run HyP3 job right now and circle back later to processing archived products.

Note: We want to be able to process either failed jobs or jobs that skipped the RAiDER step -- since failed jobs will not report a product URL, we'll want to stick with taking job_id as input.

apps/api/src/hyp3_api/handlers.py
```diff
@@ -22,13 +21,7 @@ def problem_format(status, message):
...
 
 
-@draft4_format_checker.checks('uuid')
-def is_uuid(val):
-    try:
-        UUID(val, version=4)
-    except ValueError:
-        return False
-    return True
+is_uuid = draft4_format_checker.checks('uuid')(is_uuid)
```
jhkennedy, line 24: I had to move the `is_uuid`  to validation to avoid a circular import since we *need* to provide the `granules` parameter in every job spec, and I need a way to validate that a HyP3 job ID "granule" is a valid third part granule so it's okay CMR doesn't return metadata for the granule.

This should leave is_uuid in this module the same as the decorated function version.

Is there any other gotchas with `granules` we need to look out for?

job_spec/ARIA_RAIDER.yml
```diff
@@ -0,0 +1,62 @@
+ARIA_RAIDER:
+  required_parameters:
+    - granules
+  parameters:
+    granules:
```
jhkennedy, line 5: It seems weird to use granules here

jhkennedy:
> I wonder if it's worth doing as part of this PR?

I lean towards yes, though it's not required. It took more mental effort than I had available last night, so I used the `granules` parameter and followed the pattern we took with Landsat and Sentinel-2 granules.

jtherrmann (approved):

PR #2215, opened 2024-04-26
Whitelisting

jtherrmann:
Notes:

I was questioning my decision to allow `PATCH /user` to create a new user record if it does not yet exist, but <https://www.rfc-editor.org/rfc/rfc5789> states:

> If the Request-URI does not point to an existing resource, the server MAY create a new resource

Before merging this PR:
- [x] Prepare script to update the Users table

TODO:

- [x] `GET /user` should not return an error if user record does not exist; this could mean creating the user record in the database, or just returning the appropriate response
- [x] `use_case` character limit? Edit: DynamoDB allows string items to be [up to 400KB](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ServiceQuotas.html#limits-data-types), which seems like more than anyone could ever use, and I also don't necessarily want to impose a minimum length. But [here](https://swagger.io/docs/specification/data-models/data-types/#string) are the relevant OpenAPI docs in case we want to set min or max length.
- [x] Test changes in a sandbox deployment
- [x] Finish filling out API schemas
- [x] Add/update unit tests for all new/changed functionality
- [x] `TODO` items in code
- [x] Changelog entry that includes any changes we'll need to be aware of from an operational perspective, including:
    - Users table:
      - `month_of_last_credits_reset` field renamed to `_month_of_last_credit_reset`
      - `notes` field renamed to `_notes`
    - `/user` endpoints now expose all Users table fields except those prefixed by underscore
    - Allowed `application_status` values

Validate in sandbox deployment:

- [x] Users table has a new `application_status` field, which can be one of `NOT STARTED`, `PENDING`, `APPROVED`, or `REJECTED`
    - [x] Changing a user's application status is as easy as manually updating this field to one of the four allowed values
- [x] Users table `month_of_last_credits_reset` field renamed to `_month_of_last_credit_reset`
- [x] `POST /jobs`:
    - [x] Succeeds if application is `APPROVED`
    - [x] Returns `403` if user is unapproved (with an `application_status`-specific error message)
    - [x] Raises `ValueError` (exposed as internal server error) for invalid `application_status`
- [x] `GET /user` returns full user record (minus private fields)
- [x] New `PATCH /user` endpoint:
    - [x] If the application is open:
        - [x] Updates the user record with relevant EDL fields, `use_case`, and changes `application_status` to `PENDING`
        - [x] Returns full user record (minus private fields)
        - [x] Can be re-queried to update `use_case`
    - [x] Returns `403` if application is `ACCEPTED` or `REJECTED` (with an `application_status`-specific error message)
    - [x] Raises `ValueError` (exposed as internal server error) for invalid `application_status`
- [x] `GET /user`, `PATCH /user`, and `POST /jobs`:
    - [x] If user record does not exist, automatically creates user record with `remaining_credits` set to `0`, `_month_of_last_credit_reset` set to `"0"`, and `application_status` set to `NOT STARTED`
    - [x] If user record already exists and application is `APPROVED`, resets credits if needed (follows the current behavior)
    - [x] If user record already exists and application is unapproved, resets `remaining_credits` and `_month_of_last_credit_reset` to zero
- [x] `reset_credits_monthly` deployment parameter has been removed
- [x] API spec renders new changes
- [x] API error handling still works as expected after creating `dynamo.exceptions` module
- [x] `credits_per_month` removed from unapproved user record
- [x] Behavior as expected with default application status set to `NOT_STARTED` and `APPROVED`

More TODOs:

- [x] Only pull access token from cookie, store entire EDL JSON object in a private Users table field
- [x] EDL fields deleted from API spec
- [x] Less apologetic rejection message
- [x] `_reset_credits_if_needed` elif branch: not approved and remaining credits != 0
- [x] Don't repeat the rejection exception message (in `jobs.py` and `user.py`); other repeats?
- [x] `_month_of_last_credit_reset` should default to not exists rather than `"0"`
- [x] Validate that `_reset_credits_if_needed` still works after switching to `update_item`
- [x] Delete `credits_per_month` if user is not approved? (`_reset_credits_if_needed` elif branch)
- [x] Don't hard-code application status values in unit tests
- [x] Set whitelisting sandbox default application status back to `NOT_STARTED` before sprint review demo

Add/update tests for:
- [x] New `PATCH /user` endpoint
- [x] Modified response structure for `GET /user`
- [x] `apps/api/src/hyp3_api/handlers.py`:
  - [x] `post_jobs`
  - [x] `patch_user`
  - [x] `get_user`
- [x] `apps/api/src/hyp3_api/routes.py`
  - [x] `user_patch`
- [x] `lib/dynamo/dynamo/jobs.py`
  - [x] `put_jobs`
- [x] `lib/dynamo/dynamo/user.py`
  - [x] `update_user`
  - [x] `get_or_create_user`
  - [x] `_create_user`
  - [x] `_reset_credits_if_needed`


apps/main-cf.yml.j2
```diff
@@ -36,12 +36,14 @@ Parameters:
...
+  DefaultApplicationStatus:
+    Description: The default status for new user applications.
     Type: String
     AllowedValues:
-      - false
-      - true
+      - NOT_STARTED
+      - PENDING
+      - APPROVED
+      - REJECTED
```
asjohnston-asf, line 46: Can probably limit the options to `NOT_STARTED` and `APPROVED` here

jtherrmann, line 46:
```suggestion
      - NOT_STARTED
      - APPROVED
```

asjohnston-asf (approved):

PR #2268, opened 2024-05-09
Add user application form

jtherrmann:
TODO:

- [ ] Determine how to configure our [Flask app](https://github.com/ASFHyP3/hyp3/blob/develop/apps/api/src/hyp3_api/__init__.py) to allow us to move the `request_access/` directory out of `ui/swagger/`
- [ ] Better alternative to the `DISABLE_EDL_REDIRECT` environment variable? Some options:
    - Allow logging in as a real EDL user with local api?
    - If this is not feasible, is there a cleaner way to detect if we're running a local api, and skip EDL queries?
- [x] Render HTML error responses
- [x] Disallow empty form submission
- [x] Redirect to EDL
- [x] Populate form content
- [x] Header for HTML pages, with ASF and HyP3 logos (linking out to respective websites)?
- [ ] TODOs in code
- [ ] Add/update tests
- [ ] Validate in sandbox deployment in multiple browsers and OSes (including mobile)
- [ ] Changelog entry


jtherrmann:
Closed in favor of a Vertex form implementation.

PR #2285, opened 2024-05-20
Access codes for automatic user approval

jtherrmann:
Other changes not listed in the Changelog:
- Remove the `default_credits` parameter from the `dynamo.user._reset_credits_if_needed` function
- Add a `dynamo.util.current_utc_time` function

TODO:

- [x] Add/update tests
    - [x] `apps/api/src/hyp3_api/handlers.py`: `patch_user`
    - [x] `lib/dynamo/dynamo/user.py`: `update_user`
- [x] Validate changes in sandbox deployment
- [x] `TODO`s in code
- [x] Changelog


apps/api/src/hyp3_api/api-spec/openapi-spec.yml.j2
```diff
@@ -315,6 +317,12 @@ components:
       type: string
       example: I want to process data.
 
+    access_code:
+      description: Grants automatic user approval while the code remains active.
+      type: string
+      format: uuid
```
forrestfwilliams, line 323: will need to change to string

lib/dynamo/dynamo/user.py
```diff
@@ -21,34 +26,62 @@ def update_user(user_id: str, edl_access_token: str, body: dict) -> dict:
...
+                    **access_code_value
                 },
                 ReturnValues='ALL_NEW',
             )['Attributes']
         except botocore.exceptions.ClientError as e:
             if e.response['Error']['Code'] == 'ConditionalCheckFailedException':
                 raise DatabaseConditionException(f'Failed to update record for user {user_id}')
             raise
-        return user
+        return _reset_credits_if_needed(user=user, current_month=_get_current_month(), users_table=users_table)
```
forrestfwilliams, line 66:
```suggestion
        user = _reset_credits_if_needed(user=user, current_month=_get_current_month(), users_table=users_table)
        return user
```

lib/dynamo/dynamo/user.py
```diff
@@ -21,34 +26,62 @@ def update_user(user_id: str, edl_access_token: str, body: dict) -> dict:
...
     raise InvalidApplicationStatusError(user_id, application_status)
 
 
+def _validate_access_code(access_code: str) -> None:
+    access_codes_table = DYNAMODB_RESOURCE.Table(environ['ACCESS_CODES_TABLE_NAME'])
+    item = access_codes_table.get_item(Key={'access_code': access_code}).get('Item')
+
+    if item is None:
+        raise AccessCodeError(f'{access_code} is not a valid access code')
+
```
forrestfwilliams, line 80: Maybe a `not yet active` case as well?

tests/test_api/test_patch_user.py
```diff
@@ -66,3 +66,67 @@ def test_patch_rejected_user(client, tables):
 
     assert response.status_code == HTTPStatus.FORBIDDEN
     assert 'has been rejected' in response.json['detail']
+
+
```
forrestfwilliams, line 70: Should add a test case for `not yet active` as well.

lib/dynamo/dynamo/util.py
```diff
@@ -21,13 +21,17 @@ def get_request_time_expression(start, end):
...
 
-def format_time(time: datetime):
+def format_time(time: datetime) -> str:
     if time.tzinfo is None:
         raise ValueError(f'missing tzinfo for datetime {time}')
     utc_time = time.astimezone(timezone.utc)
     return utc_time.isoformat(timespec='seconds')
 
 
+def current_time() -> str:
```
forrestfwilliams, line 31: How about `current_utc_time`?

apps/api/src/hyp3_api/api-spec/openapi-spec.yml.j2
jtherrmann: FYI @forrestfwilliams, we had discussed setting a minimum length for `access_code`, but as far as I can tell there's no way to enforce a minimum length for the DynamoDB field, so it will be up to HyP3 operators to self-enforce a minimum length, which means there's no point in setting a minimum length in the API spec (as this would just restrict what codes users can submit).

forrestfwilliams: Yep, that's fine.

forrestfwilliams (approved):

PR #2314, opened 2024-06-04
Cut over the tibet account to the M instance family

jhkennedy:


.github/workflows/deploy-enterprise.yml
```diff
@@ -60,8 +60,10 @@ jobs:
             default_credits_per_user: 0
             default_application_status: APPROVED
             cost_profile: DEFAULT
-            job_files: job_spec/INSAR_ISCE.yml
-            instance_types: c6id.xlarge,c6id.2xlarge,c6id.4xlarge,c6id.8xlarge
+            job_files: >-
+              job_spec/ARIA_RAIDER.yml
+              job_spec/INSAR_ISCE_TEST.yml
```
jhkennedy, line 65: Note: INSAR_ISCE and INSAR_ISCE_TEST only differ by the `++omp-num-threads` argument right now, and test doesn't override the image tag

cmarshak: Shouldn't we modify that so the correct number of jobs fit?
https://github.com/ASFHyP3/hyp3/blob/develop/job_spec/INSAR_ISCE_TEST.yml#L100

Looks like we might put too many jobs on an instance without modifying this too...

jhkennedy: @cmarshak this is how `hyp3-a19-jpl-test` is configured, see:
https://github.com/ASFHyP3/hyp3/blob/develop/.github/workflows/deploy-enterprise-test.yml#L41-L58

cmarshak (approved):

.github/workflows/deploy-enterprise.yml
cmarshak: I understand now - thanks!

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * The `SRG_GSLC_CPU` job spec
 * The `SRG_GSLC_CPU` job type to the `hyp3-lavas` and `hyp3-lavas-test` HyP3 deployments
 
+### Changed
+- A `hyp3-tibet-jpl` deployment now uses the `m6id[n]` instance families and includes the `ARIA_RAIDER` job spec
```
jtherrmann, line 14:
```suggestion
- The `hyp3-tibet-jpl` deployment now uses the `m6id[n]` instance families and includes the `ARIA_RAIDER` job spec
```

jtherrmann (approved):

PR #2385, opened 2024-08-19
fix INSAR_MULTI_BURST example scene names

jtherrmann:


job_spec/INSAR_MULTI_BURST.yml
```diff
@@ -11,7 +11,7 @@ INSAR_MULTI_BURST:
         maxItems: 50
```
forrestfwilliams, line 11: 50 seems like a very large number that could tax our systems. What about setting it to 30? This would be roughly the size of a standard SLC.

job_spec/INSAR_MULTI_BURST.yml
```diff
@@ -11,7 +11,7 @@ INSAR_MULTI_BURST:
...
         example:
           - S1_136231_IW2_20200604T022312_VV_7C85-BURST
-          - S1_136232_IW2_20200604T022312_VV_7C85-BURST
+          - S1_136232_IW2_20200604T022315_VV_7C85-BURST
         items:
           description: Name of the reference Sentinel-1 SLC IW burst granule to process
```
forrestfwilliams, line 16:
```suggestion
          description: Name of the reference Sentinel-1 SLC IW burst granule(s) to process
```

job_spec/INSAR_MULTI_BURST.yml
```diff
@@ -28,7 +28,7 @@ INSAR_MULTI_BURST:
         maxItems: 50
         example:
           - S1_136231_IW2_20200616T022313_VV_5D11-BURST
-          - S1_136232_IW2_20200616T022313_VV_5D11-BURST
+          - S1_136232_IW2_20200616T022316_VV_5D11-BURST
         items:
           description: Name of the secondary Sentinel-1 SLC IW burst granule to process
```
forrestfwilliams, line 33:
```suggestion
          description: Name of the secondary Sentinel-1 SLC IW burst granule(s) to process
```

job_spec/INSAR_MULTI_BURST.yml
jtherrmann: Yeah I think @asjohnston-asf had thrown out 50 as a ballpark number, but 30 seems fine too. We can always increase later.

AndrewPlayer3 (approved):

PR #2386, opened 2024-08-20
Moved INSAR_MULTI_BURST to INSAR_ISCE_BURST

AndrewPlayer3:
INSAR_ISCE_BURST now accepts reference and secondary parameters for single/multi-burst jobs, rather than having a separate multi-burst job.

The burst validation functions now accept an arbitrary number of granules, rather than just a reference and secondary.


CHANGELOG.md
```diff
@@ -5,10 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 
 
-## [7.8.0]
+## [8.0.0]
 
-### Added
-- Added `INSAR_MULTI_BURST` job type to `hyp3-test` that can merge adjacent bursts into an SLC and then perform InSAR.
+### Changed
+- `INSAR_ISCE_BURST` now has the ability to process multiple adjacent bursts into one interferogram.
+- `INSAR_ISCE_BURST` now uses `reference` and `secondary` parameters which each accept a list of 1 or more scene names.
```
jtherrmann, line 12: My understanding from this morning's conversation with @forrestfwilliams was that we would be releasing the parameter name change first, and then enabling multi-burst functionality in a subsequent release. I'm open to correction.
```suggestion
- `INSAR_ISCE_BURST` now accepts `reference` and `secondary` parameters which are each a list containing a single granule name. In a future release, `INSAR_ISCE_BURST` will support processing multiple adjacent bursts into one interferogram, and each parameter will be a list of one or more granule names.
```

CHANGELOG.md
```diff
@@ -5,10 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+
+# Removed
```
jtherrmann, line 14:
```suggestion
### Removed
```

job_spec/INSAR_ISCE_BURST.yml
```diff
@@ -1,23 +1,39 @@
...
   parameters:
-    granules:
+    reference:
       default:  '""'
       api_schema:
         type: array
-        minItems: 2
-        maxItems: 2
+        minItems: 1
+        maxItems: 30
```
jtherrmann, line 11:
```suggestion
        maxItems: 1
```

job_spec/INSAR_ISCE_BURST.yml
```diff
@@ -1,23 +1,39 @@
...
+          minLength: 43
+          maxLength: 43
+          example: 
+            - S1_136231_IW2_20200604T022312_VV_7C85-BURST
+    secondary:
+      default:  '""'
+      api_schema:
+        type: array
+        minItems: 1
+        maxItems: 30
```
jtherrmann, line 27:
```suggestion
        maxItems: 1
```

job_spec/INSAR_ISCE_BURST.yml
```diff
@@ -1,23 +1,39 @@
...
         example:
           - S1_136231_IW2_20200604T022312_VV_7C85-BURST
+        items:
+          description: Name of the reference Sentinel-1 SLC IW burst granule to process
+          type: string
+          pattern: '^((S1_\d{6}_IW\d_\d{8}T\d{6}_[VH]{2}_([\dA-F]){4}-BURST)($|[ \t]+))+$'
+          minLength: 43
+          maxLength: 43
+          example: 
+            - S1_136231_IW2_20200604T022312_VV_7C85-BURST
```
jtherrmann, line 21:
```suggestion
          example: S1_136231_IW2_20200604T022312_VV_7C85-BURST
```

apps/api/src/hyp3_api/validation.py
```diff
@@ -77,21 +77,24 @@ def check_dem_coverage(granule_metadata):
...
 
 def check_same_burst_ids(granule_metadata):
-    ref_burst_id, sec_burst_id = [granule['name'].split('_')[1] for granule in granule_metadata]
-    if ref_burst_id != sec_burst_id:
+    burst_ids = [granule['name'].split('_')[1] for granule in granule_metadata]
+    not_matching = [burst_id for burst_id in set(burst_ids) if burst_ids.count(burst_id) == 1]
+    if not_matching:
         raise GranuleValidationError(
-            f'The requested scenes do not have the same burst ID: {ref_burst_id} and {sec_burst_id}'
+            f'The requests scenes have burst IDs with no matching pairs: {not_matching}.'
```
jtherrmann, line 84: Typo?
```suggestion
            f'The requested scenes have burst IDs with no matching pairs: {not_matching}.'
```

apps/api/src/hyp3_api/validation.py
```diff
@@ -77,21 +77,24 @@ def check_dem_coverage(granule_metadata):
 
 
 def check_same_burst_ids(granule_metadata):
-    ref_burst_id, sec_burst_id = [granule['name'].split('_')[1] for granule in granule_metadata]
-    if ref_burst_id != sec_burst_id:
+    burst_ids = [granule['name'].split('_')[1] for granule in granule_metadata]
+    not_matching = [burst_id for burst_id in set(burst_ids) if burst_ids.count(burst_id) == 1]
+    if not_matching:
```
jtherrmann, line 82: Would this implementation perhaps become simpler and easier to understand if the validator function accepted `reference` and `secondary` parameters?

CHANGELOG.md
forrestfwilliams: Yep @jtherrmann that's correct. *Technically* the plugin is capable of handling multiple reference/secondary scenes at this point, but we're limiting the number of input scenes to one each for now to disable the functionality. I like this re-write of the changelog.

apps/api/src/hyp3_api/validation.py
forrestfwilliams: Yeah I think that would be better.

AndrewPlayer3: Yes but all of the direct validator functions take granule_metadata because of validate_jobs. We could re-write validate_jobs and get_granules, but that would be at the cost of consistency across how we do validation.

apps/api/src/hyp3_api/validation.py
```diff
@@ -77,21 +77,24 @@ def check_dem_coverage(granule_metadata):
 
 
 def check_same_burst_ids(granule_metadata):
-    ref_burst_id, sec_burst_id = [granule['name'].split('_')[1] for granule in granule_metadata]
-    if ref_burst_id != sec_burst_id:
+    burst_ids = [granule['name'].split('_')[1] for granule in granule_metadata]
+    not_matching = [burst_id for burst_id in set(burst_ids) if burst_ids.count(burst_id) != 2]
```
jtherrmann, line 81: My concern with this approach is that we're not actually checking if the two lists of granules were passed in the correct order, we're just checking if each burst ID appears twice in the flattened list of granules. We *could* split the list in half to get the reference and secondary lists, but I'm not sure if we can rely on the order of the granules being preserved after they get passed through `get_cmr_metadata`. I feel like the correct approach would be to refactor our validation approach to be more generalized.

apps/api/src/hyp3_api/validation.py
```diff
@@ -70,31 +70,38 @@ def check_granules_exist(granules, granule_metadata):
         raise GranuleValidationError(f'Some requested scenes could not be found: {", ".join(not_found_granules)}')
 
 
-def check_dem_coverage(granule_metadata):
+def check_dem_coverage(_, granule_metadata):
```
AndrewPlayer3, line 73:
```suggestion
def check_dem_coverage(job, granule_metadata):
```
I couldn't decide which of these is better. I ended up going with `_` because it communicates that the variable is unused.

apps/api/src/hyp3_api/validation.py
```diff
@@ -70,31 +70,38 @@ def check_granules_exist(granules, granule_metadata):
...
-    ref_burst_id, sec_burst_id = [granule['name'].split('_')[1] for granule in granule_metadata]
-    if ref_burst_id != sec_burst_id:
+def check_same_burst_ids(job, _):
+    ref_ids = [ref.split('_')[1] for ref in job['job_parameters'].get('reference', [])]
+    sec_ids = [sec.split('_')[1] for sec in job['job_parameters'].get('secondary', [])]
+    if ref_ids != sec_ids:
         raise GranuleValidationError(
-            f'The requested scenes do not have the same burst ID: {ref_burst_id} and {sec_burst_id}'
+            'The requested scenes have burst IDs with no matching pairs.'
+        )
```
AndrewPlayer3, line 85: Do we think its worth extra code here to communicate which burst pairs don't match?

tests/test_api/test_validation.py
```diff
@@ -84,75 +84,215 @@ def test_check_dem_coverage():
...
-    with raises(validation.GranuleValidationError, match=r'.*do not have the same polarization.*'):
-        validation.check_valid_polarizations(different_polarizations)
-    with raises(validation.GranuleValidationError, match=r'.*Only VV and HH.*'):
-        validation.check_valid_polarizations(unsupported_polarizations)
+def test_check_valid_polarizations():
+    valid_cases = [
+        [
+            {
+                'name': 'S1_136231_IW2_20200604T022312_VV_7C85-BURST'
+            },
```
AndrewPlayer3, line 207: The other tests didn't collapse these braces, so I didn't, but I would really like to.

apps/api/src/hyp3_api/validation.py
```diff
@@ -70,31 +70,38 @@ def check_granules_exist(granules, granule_metadata):
...
     if bad_granules:
         raise GranuleValidationError(f'Some requested scenes do not have DEM coverage: {", ".join(bad_granules)}')
 
 
-def check_same_burst_ids(granule_metadata):
-    ref_burst_id, sec_burst_id = [granule['name'].split('_')[1] for granule in granule_metadata]
-    if ref_burst_id != sec_burst_id:
+def check_same_burst_ids(job, _):
+    ref_ids = [ref.split('_')[1] for ref in job['job_parameters'].get('reference', [])]
+    sec_ids = [sec.split('_')[1] for sec in job['job_parameters'].get('secondary', [])]
```
jtherrmann, line 81: Since these are required params, no need to use `get`.
```suggestion
    ref_ids = [ref.split('_')[1] for ref in job['job_parameters']['reference']
    sec_ids = [sec.split('_')[1] for sec in job['job_parameters']['secondary']
```

apps/api/src/hyp3_api/validation.py
jtherrmann: I think that would be nice for the user, considering they may be submitting dozens of pairs. How about just loop through the pairs and report the first pair that doesn't match?

jtherrmann: Also, I find the wording of "The requested scenes have burst IDs with no matching pairs" confusing, but no need to revise it if it's going to be replaced by an error message for the specific pair that doesn't match.

apps/api/src/hyp3_api/validation.py
```diff
@@ -70,31 +70,38 @@ def check_granules_exist(granules, granule_metadata):
...
+        raise GranuleValidationError(
+            'The requested scenes have more than 1 pair with the same burst ID.'
         )
 
 
-def check_valid_polarizations(granule_metadata):
-    ref_polarization, sec_polarization = [granule['name'].split('_')[4] for granule in granule_metadata]
-    if ref_polarization != sec_polarization:
+def check_valid_polarizations(_, granule_metadata):
+    polarizations = set(granule['name'].split('_')[4] for granule in granule_metadata)
```
jtherrmann, line 93: Let's use the granules from `job` unless we specifically need the extra information from `granule_metadata`. This makes it more obvious which validators actually rely on CMR metadata, which may help use with future refactoring efforts.
```suggestion
def check_valid_polarizations(job, _):
    polarizations = set(granule.split('_')[4] for granule in get_granules([job]))
```

tests/test_api/test_validation.py
jtherrmann: I say go for it, there's no reason to preserve awkward style :)

tests/test_api/test_validation.py
```diff
@@ -84,75 +84,215 @@ def test_check_dem_coverage():
...
-    validation.check_valid_polarizations(valid_case)
-    with raises(validation.GranuleValidationError, match=r'.*do not have the same polarization.*'):
-        validation.check_valid_polarizations(different_polarizations)
-    with raises(validation.GranuleValidationError, match=r'.*Only VV and HH.*'):
-        validation.check_valid_polarizations(unsupported_polarizations)
+def test_check_valid_polarizations():
+    valid_cases = [
+        [
+            {
+                'name': 'S1_136231_IW2_20200604T022312_VV_7C85-BURST'
```
jtherrmann, line 206: If my suggestion at https://github.com/ASFHyP3/hyp3/pull/2386/files#r1725857878 is accepted, then `name` will need to be replaced by `reference` and `secondary`, like you did above.

tests/test_api/test_validation.py
```diff
@@ -84,75 +84,215 @@ def test_check_dem_coverage():
...
+            },
+        ]
+    ]
+    for valid_case in valid_cases:
+        validation.check_valid_polarizations({}, valid_case)
+    for (invalid_case_no_match, invalid_case_unsupported) in zip(invalid_cases_no_match, invalid_cases_unsupported):
+        with raises(validation.GranuleValidationError, match=r'.*need to have the same polarization*'):
+            validation.check_valid_polarizations({}, invalid_case_no_match)
+        with raises(validation.GranuleValidationError, match=r'.*currently supported*'):
+            validation.check_valid_polarizations({}, invalid_case_unsupported)
```
jtherrmann, line 295: Not a bad idea to use `zip`, but I think two separate `for` loops is more readable, like you did above.

tests/test_api/test_validation.py
```diff
@@ -230,6 +370,10 @@ def test_validate_jobs():
     granule_with_dem_coverage = 'S1A_IW_SLC__1SSV_20150621T120220_20150621T120232_006471_008934_72D8'
     granule_without_dem_coverage = 'S1A_IW_GRDH_1SDV_20201219T222530_20201219T222555_035761_042F72_8378'
 
+    valid_ref_burst = 'S1_136231_IW2_20200604T022312_VV_7C85-BURST'
+    valid_sec_burst = 'S1_136231_IW2_20200616T022313_VV_5D11-BURST'
+    invalid_sec_burst = 'S1_136232_IW2_20200616T022313_VV_5D11-BURST'
```
jtherrmann, line 375: I forgot how `test_validate_jobs` works; what specifically makes `invalid_sec_burst` an invalid granule?

job_spec/INSAR_ISCE_BURST.yml
```diff
@@ -1,23 +1,38 @@
...
+        maxItems: 1
         example:
           - S1_136231_IW2_20200604T022312_VV_7C85-BURST
-          - S1_136231_IW2_20200616T022313_VV_5D11-BURST
         items:
-          description: Name of the Sentinel-1 SLC IW burst granule to process
+          description: Name of the reference Sentinel-1 SLC IW burst granule to process
           type: string
-          pattern: "^S1_\\d{6}_IW.*-BURST"
+          pattern: '^((S1_\d{6}_IW\d_\d{8}T\d{6}_[VH]{2}_([\dA-F]){4}-BURST)($|[ \t]+))+$'
```
jtherrmann, line 17: No longer using a space-separated list.
```suggestion
          pattern: '^S1_\d{6}_IW\d_\d{8}T\d{6}_[VH]{2}_([\dA-F]){4}-BURST$'
```

job_spec/INSAR_ISCE_BURST.yml
```diff
@@ -1,23 +1,38 @@
...
+    reference:
       default:  '""'
       api_schema:
         type: array
-        minItems: 2
-        maxItems: 2
+        minItems: 1
+        maxItems: 1
         example:
           - S1_136231_IW2_20200604T022312_VV_7C85-BURST
```
jtherrmann, line 13: I wonder if we need an example for the array since we also have an example for the item. Not sure what it would look like in the Swagger UI if we removed the array example.

job_spec/INSAR_ISCE_BURST.yml
```diff
@@ -1,23 +1,38 @@
...
+      api_schema:
+        type: array
+        minItems: 1
+        maxItems: 1
+        example:
+          - S1_136231_IW2_20200616T022313_VV_5D11-BURST
+        items:
+          description: Name of the secondary Sentinel-1 SLC IW burst granule to process
+          type: string
+          pattern: '^((S1_\d{6}_IW\d_\d{8}T\d{6}_[VH]{2}_([\dA-F]){4}-BURST)($|[ \t]+))+$'
```
jtherrmann, line 32:
```suggestion
          pattern: '^S1_\d{6}_IW\d_\d{8}T\d{6}_[VH]{2}_([\dA-F]){4}-BURST$'
```

job_spec/INSAR_ISCE_BURST.yml
AndrewPlayer3: We include both for all other jobs. I'm not sure if it does anything or not.

jtherrmann: Sounds good, let's leave it then.

tests/test_api/test_validation.py
AndrewPlayer3: It's not valid because it's id doesn't match the above granules id. I've made this more clear by creating a valid_burst_pair and invalid_burst_pair.

apps/api/src/hyp3_api/validation.py
```diff
@@ -70,31 +70,48 @@ def check_granules_exist(granules, granule_metadata):
...
+        error_message = (
+            f'Number of reference and secondary scenes must match, got: '
+            f'{len(ref_ids)} references and {len(sec_ids)} secondaries'
+        )
+        raise GranuleValidationError(error_message)
+    if ref_ids != sec_ids:
+        bad_id_index = [ref_ids.index(ref) for (ref, sec) in zip(ref_ids, sec_ids) if ref != sec][0]
+        raise GranuleValidationError(
+            f'Burst IDs do not match for {refs[bad_id_index]} and {secs[bad_id_index]}.'
+        )
```
jtherrmann, line 94: This is the obvious way to do it, so takes less time to understand, in my opinion.
```suggestion
    for i in range(len(ref_ids)):
        if ref_ids[i] != sec_ids[i]:
            raise GranuleValidationError(
                f'Burst IDs do not match for {refs[i]} and {secs[i]}.'
            )
```

apps/api/src/hyp3_api/validation.py
```diff
@@ -70,31 +70,48 @@ def check_granules_exist(granules, granule_metadata):
...
+    refs = job['job_parameters']['reference']
+    secs = job['job_parameters']['secondary']
+    ref_ids = [ref.split('_')[1] for ref in refs]
+    sec_ids = [sec.split('_')[1] for sec in secs]
+    if len(ref_ids) != len(sec_ids):
+        error_message = (
+            f'Number of reference and secondary scenes must match, got: '
+            f'{len(ref_ids)} references and {len(sec_ids)} secondaries'
+        )
+        raise GranuleValidationError(error_message)
```
jtherrmann, line 89:
```suggestion
        raise GranuleValidationError(
            f'Number of reference and secondary scenes must match, got: '
            f'{len(ref_ids)} references and {len(sec_ids)} secondaries'
        )
```

apps/api/src/hyp3_api/validation.py
AndrewPlayer3: I've been staring at so many list comprehensions that that somehow didn't occur to me lol

AndrewPlayer3: My only complaint would be that it runs the loop every time, but I guess these are all tiny lists so it shouldn't matter.

apps/api/src/hyp3_api/validation.py
```diff
@@ -70,31 +70,48 @@ def check_granules_exist(granules, granule_metadata):
...
+    if ref_ids != sec_ids:
+        bad_id_index = [ref_ids.index(ref) for (ref, sec) in zip(ref_ids, sec_ids) if ref != sec][0]
+        raise GranuleValidationError(
+            f'Burst IDs do not match for {refs[bad_id_index]} and {secs[bad_id_index]}.'
+        )
+    if len(set(ref_ids)) != len(ref_ids):
+        duplicate_pair_id = [ref_id for ref_id in ref_ids if ref_ids.count(ref_id) > 1][0]
```
jtherrmann, line 95: Note that this is an O(n^2) operation as written. Even if we allowed 100 granule pairs, that would only be 10,000 iterations, so not a real performance concern, but it seems like bad style. I don't see an immediately obvious best way to re-write it, though. Maybe something with [`collections.Counter`](https://docs.python.org/3/library/collections.html#collections.Counter). I'll think more about it.

apps/api/src/hyp3_api/validation.py
jtherrmann: But keep in mind that `ref_ids != sec_ids` is also a looping operation, as is `ref_ids.index(ref)`, plus the list comprehension. So you've actually got three loops here :)

apps/api/src/hyp3_api/validation.py
jtherrmann: To clarify, it's O(n^2) because it calls `ref_ids.count(ref_id)` (which is a looping operation) for every element in the list. So it's effectively a nested loop.

AndrewPlayer3: I think I could do a filter() and then use next() so that it stops at the first duplicate pair. It would still be O(n^2) worst case, but probably much better on average. 

Like
```python
filt = lambda ref_id: ref_ids.count(ref_id) > 1
first_dup = next(filter(filt, ref_ids))
```

Although, this might actually be slower because of the function calls. Fun to think about anyway.

jtherrmann: Meh, I think it's readable enough as-is.

AndrewPlayer3: Ope, sounds good.

apps/api/src/hyp3_api/validation.py
```diff
@@ -70,31 +70,47 @@ def check_granules_exist(granules, granule_metadata):
...
+            f'Number of reference and secondary scenes must match, got: '
+            f'{len(ref_ids)} references and {len(sec_ids)} secondaries'
+        )
+    for i in range(len(ref_ids)):
+        if ref_ids[i] != sec_ids[i]:
+            raise GranuleValidationError(
+                f'Burst IDs do not match for {refs[i]} and {secs[i]}.'
+            )
+    if len(set(ref_ids)) != len(ref_ids):
+        duplicate_pair_id = [ref_id for ref_id in ref_ids if ref_ids.count(ref_id) > 1][0]
```
jtherrmann, line 95: Just a small improvement, but might as well.
```suggestion
        duplicate_pair_id = next(ref_id for ref_id in ref_ids if ref_ids.count(ref_id) > 1)
```

jtherrmann (approved):

jtherrmann:
The snyk failure was not introduced by this PR.

PR #2389, opened 2024-08-23
Multi-burst functionality

jtherrmann:
TODO:

- [x] Changelog entry


apps/set-batch-overrides/src/set_batch_overrides.py
```diff
@@ -0,0 +1,86 @@
...
+            return INSAR_ISCE_BURST_MEMORY_8G
+        if bursts < 31:
+            return INSAR_ISCE_BURST_MEMORY_16G
+    raise ValueError(f'No memory value for {bursts} bursts and {looks} looks')
+
+
+def lambda_handler(event: dict, _) -> dict:
+    job_type, job_parameters = event['job_type'], event['job_parameters']
+
+    # TODO rename to INSAR_ISCE_BURST after the two burst types are merged
```
jtherrmann, line 68:
```suggestion
    # TODO https://github.com/ASFHyP3/hyp3/issues/2442 rename to INSAR_ISCE_BURST after the two burst types are merged
```

tests/test_api/test_validation.py
```diff
@@ -270,6 +423,21 @@ def test_validate_jobs():
             'job_type': 'ARIA_RAIDER',
             'job_parameters': {}
         },
+        {
+            # TODO rename to INSAR_ISCE_BURST after the two burst types are merged
```
jtherrmann, line 427:
```suggestion
            # TODO https://github.com/ASFHyP3/hyp3/issues/2442 rename to INSAR_ISCE_BURST after the two burst types are merged
```

tests/test_api/test_validation.py
```diff
@@ -270,6 +423,21 @@ def test_validate_jobs():
...
+            'job_type': 'INSAR_ISCE_MULTI_BURST',
+            'job_parameters': {
+                'reference': [valid_burst_pair[0]],
+                'secondary': [valid_burst_pair[1]]
+            }
+        },
+        {
+            # TODO remove this test case after the two burst types are merged
```
jtherrmann, line 435:
```suggestion
            # TODO https://github.com/ASFHyP3/hyp3/issues/2442 remove this test case after the two burst types are merged
```

tests/test_api/test_validation.py
```diff
@@ -294,3 +462,28 @@ def test_validate_jobs():
     ]
     with raises(validation.GranuleValidationError):
         validation.validate_jobs(jobs)
+
+    jobs = [
+        {
+            # TODO rename to INSAR_ISCE_BURST after the two burst types are merged
```
jtherrmann, line 468:
```suggestion
            # TODO https://github.com/ASFHyP3/hyp3/issues/2442 rename to INSAR_ISCE_BURST after the two burst types are merged
```

tests/test_api/test_validation.py
```diff
@@ -294,3 +462,28 @@ def test_validate_jobs():
...
+            'job_parameters': {
+                'reference': [invalid_burst_pair[0]],
+                'secondary': [invalid_burst_pair[1]]
+            }
+        }
+    ]
+    with raises(validation.GranuleValidationError):
+        validation.validate_jobs(jobs)
+
+    # TODO remove this test case after the two burst types are merged
```
jtherrmann, line 479:
```suggestion
    # TODO https://github.com/ASFHyP3/hyp3/issues/2442 remove this test case after the two burst types are merged
```

tests/test_set_batch_overrides.py
```diff
@@ -0,0 +1,275 @@
...
+    INSAR_ISCE_BURST_MEMORY_8G,
+    RTC_GAMMA_10M_MEMORY,
+    WATER_MAP_10M_MEMORY,
+    lambda_handler,
+)
+
+
+def mock_insar_isce_burst_job(looks: str, bursts: int) -> dict:
+    return {
+        # TODO rename to INSAR_ISCE_BURST after the two burst types are merged
```
jtherrmann, line 19:
```suggestion
        # TODO https://github.com/ASFHyP3/hyp3/issues/2442 rename to INSAR_ISCE_BURST after the two burst types are merged
```

apps/api/src/hyp3_api/validation.py
```diff
@@ -119,7 +136,21 @@ def get_multipolygon_from_geojson(input_file):
...
-def validate_jobs(jobs):
+# TODO https://github.com/ASFHyP3/hyp3/issues/2442 remove this function after two burst types are merged
+def convert_single_burst_jobs(jobs: list[dict]) -> list[dict]:
+    jobs = deepcopy(jobs)
+    for job in jobs:
+        if job['job_type'] == 'INSAR_ISCE_BURST':
+            job_parameters = job['job_parameters']
+            job_parameters['reference'], job_parameters['secondary'] = (
+                [granule] for granule in job_parameters.pop('granules')
+            )
```
AndrewPlayer3, line 147: I think the intention of the code is more clear if written like this suggestion. Also, a doc-string or comment that gives insight as to why this function is necessary would be great for future reference, since I'm not sure when the job type merging will happen.

```suggestion
            ref, sec = tuple(job_parameters.pop('granules'))
            job_parameters['reference'], job_parameters['secondary'] = [ref], [sec]
```

apps/set-batch-overrides/src/set_batch_overrides.py
```diff
@@ -0,0 +1,86 @@
...
+        if bursts < 21:
+            return INSAR_ISCE_BURST_MEMORY_16G
+        if bursts < 31:
+            return INSAR_ISCE_BURST_MEMORY_32G
+    if looks == '20x4':
+        if bursts < 23:
+            return INSAR_ISCE_BURST_MEMORY_8G
+        if bursts < 31:
+            return INSAR_ISCE_BURST_MEMORY_16G
+    raise ValueError(f'No memory value for {bursts} bursts and {looks} looks')
```
AndrewPlayer3, line 62: This probably won't change too much, so it may not matter, but this would be more extensible and clean if the limits were in a dict. Something like:

```python
if looks == '5x1':
  burst_limits_5x1 = {
      2: INSAR_ISCE_BURST_MEMORY_8G,
      4: INSAR_ISCE_BURST_MEMORY_16G, 
      11: INSAR_ISCE_BURST_MEMORY_32G,
      25: INSAR_ISCE_BURST_MEMORY_64G,
      31: INSAR_ISCE_BURST_MEMORY_128G
  }
  mem_req_key = min(burst_limits_5x1, key=lambda x: abs(x-bursts))
  return burst_limits_5x1[mem_req_key]
if looks == '10x2':
  ...
``

File: job_spec/INSAR_ISCE_MULTI_BURST.yml [hunk 9]
```diff
@@ -0,0 +1,87 @@
...
+    - secondary
+  parameters:
+    reference:
+      default:  '""'
+      api_schema:
+        type: array
+        minItems: 1
+        maxItems: 30
+        example:
+          - S1_136231_IW2_20200604T022312_VV_7C85-BURST
```
AndrewPlayer3, line 13: The examples should probably have multiple bursts for this job type.

AndrewPlayer3 (commented):
Looks pretty good, just a couple comments.

job_spec/INSAR_ISCE_MULTI_BURST.yml
jtherrmann: Agreed. Perhaps we just add that as a TODO for when we release multi-burst InSAR to EDC prod?

apps/set-batch-overrides/src/set_batch_overrides.py
jtherrmann: Yeah, I had a full implementation of something similar, using a data structure (I think a list of intervals) rather than a bunch of `if` statements, but in the end I decided it was actually less readable and flexible. Feel free to take a stab at a refactor if you're inspired, but I'm content with the current version.

job_spec/INSAR_ISCE_MULTI_BURST.yml
jtherrmann: I added a TODO.

apps/api/src/hyp3_api/validation.py
jtherrmann: Agreed, that's more readable. Although, you can remove the cast to `tuple`, as the value unpacking will work without it.

jtherrmann: Also, if you follow the link to the GitHub issue included in the comment above this function, it explains the context for why it's necessary.

jtherrmann: I implemented your suggestion.

PR #2390, opened 2024-08-26
Per-job Instance Types and AMIs

AndrewPlayer3:
TODO verify in `hyp3-lavas-test` (this list can be moved to the release PR after this PR is merged):

- [ ] All new resources are created with the expected parameters
- [ ] All modified resources are updated appropriately
- [ ] All Lambda functions continue to function as expected for the default Batch cluster and the new one
  - [ ] `handle_batch_event`
  - [ ] `scale_cluster` (including the new Lambda function)
- [ ] Step function submits jobs to the appropriate job queue
- [ ] The `SRG_GSLC` job gets submitted with the requested GPU resources and the `--gpu` option


apps/main-cf.yml.j2
```diff
@@ -180,7 +188,11 @@ Resources:
     Type: AWS::CloudFormation::Stack
     Properties:
       Parameters:
-        JobQueueArn: !GetAtt Cluster.Outputs.JobQueueArn
+        SharedJobQueueArn: !GetAtt Cluster.Outputs.SharedJobQueueArn
+        {% for job_type, job_spec in job_types.items() if 'Shared' not in job_spec['compute_environment']['name'] %}
```
jtherrmann, line 192: Any reason not to use a more strict comparison like this?
```suggestion
        {% for job_type, job_spec in job_types.items() if job_spec['compute_environment']['name'] != 'Shared' %}
```

jtherrmann: There are several places where this change could be made.

apps/compute-cf.yml.j2
```diff
@@ -63,7 +73,70 @@ Resources:
...
+      SchedulingPolicyArn: !Ref SchedulingPolicy
+
+  SchedulingPolicy:
+    Type: AWS::Batch::SchedulingPolicy
+
+  {% for job_type, job_spec in job_types.items() if 'Shared' not in job_spec['compute_environment']['name'] %}
+  {% set env = job_spec['compute_environment'] %}
+  {% set name = env['name'] %}
+  {% set instance_types = env['instance_types'] if 'instance_types' in env else ['!Ref InstanceTypes'] %}
+  {% set ami_id = env['ami_id'] if 'ami_id' in env else '!Ref AmiId' %}
```
jtherrmann, line 114: You might be able to use the `dict.get` method here, not sure if Jinja supports it.

apps/compute-cf.yml.j2
```diff
@@ -63,7 +73,70 @@ Resources:
...
+            MIME-Version: 1.0
+
+            --==BOUNDARY==
+            Content-Type: text/cloud-boothook; charset="us-ascii"
+
+            #!/bin/bash
+            cloud-init-per instance mkfs_ssd mkfs.ext4 /dev/nvme1n1
+            mount /dev/nvme1n1 /var/lib/docker
+
+            {% for command in job_spec['compute_environment']['user_data_commands'] %}
```
jtherrmann, line 133:
```suggestion
            {% for command in env['user_data_commands'] %}
```

apps/compute-cf.yml.j2
```diff
@@ -63,7 +73,70 @@ Resources:
...
+            {{ command }}
+            {% endfor %}
+
+            --==BOUNDARY==--
```
jtherrmann, line 137: We might be able to remove the entire custom launch template now that we don't have an immediate need for custom user data commands.

apps/compute-cf.yml.j2
```diff
@@ -63,7 +73,70 @@ Resources:
 
             --==BOUNDARY==--
 
-  ComputeEnvironment:
+  SharedComputeEnvironment:
```
jtherrmann, line 76: I think we'll actually want to avoid renaming existing resources, as this will probably cause them to be deleted and re-created during the stack deployment, which could disrupt running jobs. (I agree that the `Shared` prefix makes it more readable, but probably not worth the operational disruption.)

apps/scale-cluster/scale-cluster-cf.yml.j2
```diff
@@ -126,3 +141,44 @@ Resources:
...
+      MaximumRetryAttempts: 0
+
+  {{ name }}EventPermission:
+    Type: AWS::Lambda::Permission
+    Properties:
+      FunctionName: !GetAtt  {{ name }}Lambda.Arn
+      Action: lambda:InvokeFunction
+      Principal: events.amazonaws.com
+      SourceArn: !GetAtt Schedule.Arn
+  {% endfor %}
```
jtherrmann, line 184: Newline.
```suggestion
  {% endfor %}

```

apps/compute-cf.yml.j2
```diff
@@ -85,17 +95,58 @@ Resources:
         Tags:
           Name: !Ref AWS::StackName
 
+  JobQueue:
```
jtherrmann, line 98: We'll want to keep the original name.
```suggestion
  BatchJobQueue:
```

apps/compute-cf.yml.j2
```diff
@@ -27,7 +27,17 @@ Outputs:
     Value: !Ref ComputeEnvironment
 
   JobQueueArn:
-    Value: !Ref BatchJobQueue
+    Value: !Ref JobQueue
```
jtherrmann, line 30:
```suggestion
    Value: !Ref BatchJobQueue
```

apps/scale-cluster/scale-cluster-cf.yml.j2
```diff
@@ -126,3 +141,44 @@ Resources:
...
       SourceArn: !GetAtt Schedule.Arn
+
+  {% for job_type, job_spec in job_types.items() if 'Default' != job_spec['compute_environment']['name'] %}
+  {% set name = job_spec['compute_environment']['name'] %}
+  {{ name }}Lambda:
+    Type: AWS::Lambda::Function
+    Properties:
+      Environment:
+        Variables:
+          COMPUTE_ENVIRONMENT_ARN: !Ref {{name }}ComputeEnvironmentArn
```
jtherrmann, line 152:
```suggestion
          COMPUTE_ENVIRONMENT_ARN: !Ref {{ name }}ComputeEnvironmentArn
```

apps/scale-cluster/scale-cluster-cf.yml.j2
jtherrmann, line 1: If we're adding new lambda functions as part of this template, then we'll also need a separate `LogGroup` resource for each lambda.

apps/workflow-cf.yml.j2
```diff
@@ -5,6 +5,11 @@ Parameters:
   JobQueueArn:
     Type: String
 
+  {% for job_type, job_spec in job_types.items() if 'Default' != job_spec['compute_environment']['name']%}
```
jtherrmann, line 8:
```suggestion
  {% for job_type, job_spec in job_types.items() if 'Default' != job_spec['compute_environment']['name'] %}
```

apps/workflow-cf.yml.j2
```diff
@@ -60,6 +66,8 @@ Resources:
             Value: "{{ task['vcpu'] }}"
           - Type: MEMORY
             Value: "{{ task['memory'] }}"
+          - Type: GPU
+            Value: "{{ task['gpu'] if 'gpu' in task else 0 }}"
```
jtherrmann, line 70: Probably better to just not set this value at all if the task doesn't require it.
```suggestion
          {% if 'gpu' in task %}
          - Type: GPU
            Value: "{{ task['gpu'] }}"
          {% endif %}
```

apps/main-cf.yml.j2
```diff
@@ -154,6 +154,10 @@ Resources:
     Properties:
       Parameters:
         ComputeEnvironmentArn: !GetAtt Cluster.Outputs.ComputeEnvironmentArn
+        {% for job_type, job_spec in job_types.items() if 'Default' != job_spec['compute_environment']['name'] %}
```
jtherrmann, line 157: I think I'd actually prefer to have the default be to leave `compute_environment` out of the job spec entirely, and then these loops all look like:
```suggestion
        {% for job_type, job_spec in job_types.items() if 'compute_environment' in job_spec %}
```
which just keeps the job spec slightly less cluttered. But I could probably be convinced otherwise if there's a compelling reason to always include it in the job spec.

jtherrmann:
This implementation is slightly different from how I would interpret the acceptance criteria for <https://asfdaac.atlassian.net/browse/TOOL-2870>, which specifies that there should be a "deployment-specified gpu ami and instance type fleet" and a "deployment-specified cpu ami and instance type fleet". Instead, this implementation keeps a deployment-specified default ami and instance type fleet, and allows overriding those values per job spec. This is probably more flexible so perhaps better than what was written for the acceptance criteria, but we should get buy-in from at least @forrestfwilliams on this approach.

jtherrmann:
Also note the bolded text from the Jira issue:

> **This will have implications for our current cost control strategy.**

I believe this was originally raised as a potential concern by @asjohnston-asf, and we put it in the Jira issue to remind ourselves to consider it more carefully. So we should discuss how that relates to this PR.

apps/step-function.json.j2
```diff
@@ -207,7 +207,9 @@
       "Parameters": {
         "JobDefinition": "{{ '${'+ snake_to_pascal_case(task['name']) + '}' }}",
         "JobName.$": "$.job_id",
-        "JobQueue": "${JobQueueArn}",
+        {% set name = job_spec['compute_environment']['name'] %}
+        {% set job_queue = name + 'JobQueueArn' if 'Default' != name else 'JobQueueArn' %}
+        "JobQueue": "{{ '${'+ job_queue +'}' }}",
```
jtherrmann, line 212: Note that this means the same compute environment will be used for all tasks within the job spec, which is fine for `SRG_GSLC`, but maybe we'll want to be able to specify AMI and instance types for an individual task, the same way we can specify `memory` and `vcpu` for an individual task? Not sure how difficult that would be to implement as part of this PR, so maybe not worth trying to support that until we need it.

job_spec/SRG_GSLC.yml
```diff
@@ -30,14 +30,22 @@ SRG_GSLC:
...
+      memory: 30500
       secrets:
         - EARTHDATA_USERNAME
         - EARTHDATA_PASSWORD
+  compute_environment:
+    name: SrgGslc
+    instance_types:
+      - g6.2xlarge
+    # Image ID for: /aws/service/ecs/optimized-ami/amazon-linux-2/recommended/image_id
+    ami_id: ami-0ca2af66da8e56876
```
jtherrmann, line 51: I'd suggest moving the `compute_environment` block to just before the `tasks` block (just after `cost_profiles`) to make it visually more obvious that `compute_environment` applies to the entire job spec and is not part of the individual task specification.

jhkennedy:
> which specifies that there should be a "deployment-specified gpu ami and instance type fleet" and a "deployment-specified cpu ami and instance type fleet". This is probably more flexible so perhaps better than what was written for the acceptance criteria, but we should get buy-in from at least @forrestfwilliams on this approach.

Looking ahead, we'll be moving ARIA S1 GUNW (INSAR_ISCE) jobs into production HyP3 soon, which are currently running on the C or M instance family (we found them to be the most cost-optimal). So, I think this approach is better overall as we'll have broader needs than just GPU vs CPU.

> This will have implications for our current cost control strategy.

Right now, we do cost control by limiting the max vcpus, which works well with one compute environment. Multiple compute environments will definitely reduce the effectiveness of this strategy. For the LAVAs deployment, I don't think we'll be using the EDC cost controls, so won't matter there. We can side-step the cost control worry in EDC by keeping it a single compute environment there for now, but we'll need to tackle it when we bring in GUNWs if we want to use the C/M instance families. 

I'm not sure how forward-looking we want to be overall here, but we definitely don't want to put ourselves in a corner later.

apps/main-cf.yml.j2
AndrewPlayer3: I really prefer imperative over implied. We also specify a default cost of 1.0 in all the files, which probably isn't necessary, but I find it helpful. At the very least, it makes it impossible to miss that its an option when using an existing job_spec as an example (since this isn't documented).

apps/step-function.json.j2
AndrewPlayer3: This is an interesting idea, I'm thinking about it.

apps/main-cf.yml.j2
jtherrmann: Yep, that's fair. If it were documented separately, I'd argue for leaving it out, but keeping it as a visible option in the absence of documentation is a good reason for leaving it in.

job_spec/WATER_MAP_EQ.yml
```diff
@@ -113,3 +115,4 @@ WATER_MAP_EQ:
       timeout: 36000
       vcpu: 1
       memory: 126000
+
```
jtherrmann, line 118:
```suggestion
```

job_spec/WATER_MAP.yml
```diff
@@ -203,3 +205,4 @@ WATER_MAP:
       timeout: 86400
       vcpu: 1
       memory: 126000
+
```
jtherrmann, line 208:
```suggestion
```

job_spec/SRG_GSLC.yml
```diff
@@ -24,20 +24,29 @@ SRG_GSLC:
...
         - Ref::granules
       timeout: 10800
       vcpu: 1
-      memory: 31500
+      gpu: 1
+      memory: 30500
       secrets:
         - EARTHDATA_USERNAME
         - EARTHDATA_PASSWORD
+
```
jtherrmann, line 52:
```suggestion
```

job_spec/INSAR_ISCE.yml
```diff
@@ -154,3 +156,4 @@ INSAR_ISCE:
         - RAIDER_HRES_EMAIL
         - RAIDER_HRES_API_KEY
         - RAIDER_HRES_URL
+
```
jtherrmann, line 159:
```suggestion
```

job_spec/ARIA_RAIDER.yml
```diff
@@ -52,3 +54,4 @@ ARIA_RAIDER:
         - RAIDER_HRES_EMAIL
         - RAIDER_HRES_API_KEY
         - RAIDER_HRES_URL
+
```
jtherrmann, line 57:
```suggestion
```

jtherrmann (commented):
Looks like these are just extra blank lines, not a trailing newline.

job_spec/SRG_GSLC.yml
```diff
@@ -24,20 +24,28 @@ SRG_GSLC:
...
         - --bucket
         - '!Ref Bucket'
         - --bucket-prefix
         - Ref::bucket_prefix
         - Ref::granules
       timeout: 10800
       vcpu: 1
-      memory: 31500
+      gpu: 1
+      memory: 30500
```
jtherrmann, line 48: Just want to confirm that it was intentional to decrease the memory allocation slightly?

apps/scale-cluster/scale-cluster-cf.yml.j2
```diff
@@ -79,7 +85,11 @@ Resources:
             Resource: "*"
           - Effect: Allow
             Action: batch:UpdateComputeEnvironment
-            Resource: !Ref ComputeEnvironmentArn
+            Resource: 
+            - !Ref ComputeEnvironmentArn
+            {% for job_type, job_spec in job_types.items() if 'Default' != job_spec['compute_environment']['name'] %}
+            - !Ref {{ job_spec['compute_environment']['name'] }}ComputeEnvironmentArn
+            {% endfor %}
```
jtherrmann, line 92:
```suggestion
              - !Ref ComputeEnvironmentArn
              {% for job_type, job_spec in job_types.items() if 'Default' != job_spec['compute_environment']['name'] %}
              - !Ref {{ job_spec['compute_environment']['name'] }}ComputeEnvironmentArn
              {% endfor %}
```

apps/scale-cluster/scale-cluster-cf.yml.j2
```diff
@@ -126,3 +141,50 @@ Resources:
       Action: lambda:InvokeFunction
       Principal: events.amazonaws.com
       SourceArn: !GetAtt Schedule.Arn
+
+  {% for job_type, job_spec in job_types.items() if 'Default' != job_spec['compute_environment']['name'] %}
+  {% set name = job_spec['compute_environment']['name'] %}
+  {{ name }}LogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: !Sub {{ "/aws/lambda/${"+ name +"Lambda}" }}
```
jtherrmann, line 150: Extra spaces makes it more readable for me.
```suggestion
      LogGroupName: !Sub {{ "/aws/lambda/${" + name + "Lambda}" }}
```

apps/scale-cluster/scale-cluster-cf.yml.j2
jtherrmann, line 1: Other than my two pending suggestions, this file looks good to me.

apps/workflow-cf.yml.j2
```diff
@@ -124,6 +138,9 @@ Resources:
             Action: batch:SubmitJob
             Resource:
               - !Ref JobQueueArn
+              {% for job_type, job_spec in job_types.items() if 'Default' != job_spec['compute_environment']['name']%}
```
jtherrmann, line 141:
```suggestion
              {% for job_type, job_spec in job_types.items() if 'Default' != job_spec['compute_environment']['name'] %}
```

apps/workflow-cf.yml.j2
jtherrmann, line 1: Other than my one suggestion, this file looks good to me.

apps/step-function.json.j2
jtherrmann, line 212:
```suggestion
        "JobQueue": "{{ '${' + job_queue + '}' }}",
```

apps/step-function.json.j2
jtherrmann, line 1: Other than my one suggestion, this file looks good to me.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [7.8.0]
+
+### Added
+- Support for per-job specification of instance types and AMIs.
```
jtherrmann, line 11:
```suggestion
- Allow overriding certain AWS Batch compute environment parameters (including instance types and AMI) within a job spec.
- Allow job spec tasks to require GPU resource requirements.
```

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [7.8.0]
+
+### Added
+- Allow overriding certain AWS Batch compute environment parameters (including instance types and AMI) within a job spec.
+- Allow job spec tasks to require GPU resource requirements.
```
jtherrmann, line 12:
```suggestion
- Allow job spec tasks to require GPU resources.

### Changed
- The `SRG_GSLC` job type now runs within a GPU environment.
```

jtherrmann (commented):
All of the code changes look good, but I still want to render the templates locally and review the changes.

jtherrmann (approved):
I've confirmed that the templates render locally as expected.

jtherrmann:
The snyk failure is unrelated to this PR.

PR #2397, opened 2024-08-29
Release v7.8.0

AndrewPlayer3:
TODO verify in `hyp3-lavas-test`:

- [x] All new resources are created with the expected parameters
- [x] All modified resources are updated appropriately
- [x] All Lambda functions continue to function as expected for the default Batch cluster and the new one
  - [x] `handle_batch_event`
  - [x] `scale_cluster` (for the existing cluster in edc-uat)
- [x] Step function submits jobs to the appropriate job queue
- [x] The `SRG_GSLC` job gets submitted with the requested GPU resources and the `--gpu` option

TODO:

- [x] Wait for Joe to merge his changes
- [x] Confirm that Joe has validated in `hyp3-a19-jpl-test`


jtherrmann (approved):
Approved, but let's hold off on releasing until we've confirmed that @jhkennedy has validated what he needs to in `hyp3-a19-jpl-test`

jhkennedy (approved):
Looks good! 

I've verified the behavior on the ARIA/JPL side -- in hyp3-a19-jpl-test the INSAR_ISCE jobs appropriately went to their own job queue and compete environment, while the ARIA_AUTORIFT jobs went through the default ones.

PR #2399, opened 2024-08-29
use dedicated on-demand compute environment for INSAR_ISCE jobs

jhkennedy:
This causes any deployment that includes the `INSAR_ISCE.yml` job spec to use a dedicated `Aria` compute environment with on-demand instead of spot instances. Notably, this currently effects:
* hyp3-a19-jpl
* hyp3-a19-jpl-test
* hyp3-tibet-jpl
* hyp3-nisar-jpl

In these deployments, other job specs (ARIA_AUTORIFT.yml and ARIA_RAIDER.yml) will use the default compute environment with spot instances.

### DIFF
--- a/job_spec/INSAR_ISCE.yml
+++ b/job_spec/INSAR_ISCE.yml
@@ -93,7 +93,9 @@ INSAR_ISCE:
       cost: 1.0
   validators: []
   compute_environment:
-    name: 'Default'
+    name: 'Aria'
+    allocation_type: EC2
+    allocation_strategy: BEST_FIT_PROGRESSIVE
   tasks:
     - name: ''
       image: ghcr.io/access-cloud-based-insar/dockerizedtopsapp
--- a/job_spec/INSAR_ISCE.yml
+++ b/job_spec/INSAR_ISCE.yml
@@ -93,7 +93,7 @@ INSAR_ISCE:
       cost: 1.0
   validators: []
   compute_environment:
-    name: 'Aria'
+    name: 'InsarIsceAria'
     allocation_type: EC2
     allocation_strategy: BEST_FIT_PROGRESSIVE
   tasks:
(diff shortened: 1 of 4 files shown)


job_spec/INSAR_ISCE.yml
```diff
@@ -93,7 +93,9 @@ INSAR_ISCE:
       cost: 1.0
   validators: []
   compute_environment:
-    name: 'Default'
+    name: 'Aria'
+    allocation_type: EC2
+    allocation_strategy: BEST_FIT_PROGRESSIVE
```
jhkennedy, line 98: Compute environment is rendering like:
```yaml
  AriaComputeEnvironment:
    Type: AWS::Batch::ComputeEnvironment
    Properties:
      ServiceRole: !GetAtt BatchServiceRole.Arn
      Type: MANAGED
      ComputeResources:
        Type: EC2
        AllocationStrategy: BEST_FIT_PROGRESSIVE
        MinvCpus: 0
        MaxvCpus: !Ref MaxvCpus
        InstanceTypes:
          - !Ref InstanceTypes
```
but InstanceTypes should look like:
```yaml
        InstanceTypes: !Ref InstanceTypes
```

jhkennedy: Ah, InstanceTypes is being rendered like a list, but it should be a comma-deliminated string; fixed in https://github.com/ASFHyP3/hyp3/pull/2399/commits/70952d01e0232de66a0f2d4a9cafcee0c793dbbd

job_spec/INSAR_ISCE.yml
```diff
@@ -93,7 +93,9 @@ INSAR_ISCE:
       cost: 1.0
   validators: []
   compute_environment:
-    name: 'Default'
+    name: 'Aria'
```
jtherrmann, line 96: Should we perhaps make it more explicit that this compute environment only applies to this particular job type?
```suggestion
    name: 'InsarIsceAria'
```

jhkennedy: That's a good idea!

jtherrmann (approved):

PR #2400, opened 2024-08-29
render instance types as comma-delimited list

jtherrmann:


jhkennedy (approved):
Ah right! Good catch.

PR #2447, opened 2024-10-02
Add `SRG_TIME_SERIES` job type

jtherrmann:
TODO:

- [x] Validator for relative orbit?
- [x] Decide on compute environment for `TROPOSPHERE` step in `job_spec/INSAR_ISCE.yml`
- [x] Remaining `TODO` items in `job_spec/SRG_TIME_SERIES.yml`


apps/start-execution-worker/src/start_execution_worker.py
```diff
@@ -8,24 +9,35 @@
 
 STEP_FUNCTION = boto3.client('stepfunctions')
 
+batch_params_file = Path(__file__).parent / 'batch_params_by_job_type.json'
+if batch_params_file.exists():
+    BATCH_PARAMS_BY_JOB_TYPE = json.loads(batch_params_file.read_text())
+else:
+    # Allows mocking with unittest.mock.patch
+    BATCH_PARAMS_BY_JOB_TYPE = {}
```
AndrewPlayer3, line 17: It may not be worth the effort right now, but I'm pretty sure you can patch `Path.exists`, `Path.read_text`, and `json.loads`.

AndrewPlayer3: Here are some references if you want them:
* [How to patch exists](https://stackoverflow.com/questions/48864027/how-do-i-patch-the-pathlib-path-exists-method)
* [How to patch json.loads](https://stackoverflow.com/questions/51138834/how-to-use-mock-open-with-json-load)

apps/render_cf.py
```diff
@@ -59,9 +270,33 @@ def render_costs(job_types: dict, cost_profile: str) -> None:
         json.dump(costs, f, indent=2)
 
 
+def validate_job_spec(job_type: str, job_spec: dict) -> None:
+    # Non-comprehensive job spec validator. More checks could be added if we want to be more thorough.
+
```
AndrewPlayer3, line 275: Does this cover everything that can silently fail? If so, I think its fine.
```suggestion
```

job_spec/INSAR_ISCE.yml
```diff
@@ -146,6 +140,9 @@ INSAR_ISCE:
         - --weather-model
         - Ref::weather_model
       timeout: 10800
+      compute_environment:
+        # TODO: Do we want this to use this env or Default?
```
jtherrmann, line 144:
```suggestion
```

apps/render_cf.py
```diff
@@ -11,35 +12,245 @@ def snake_to_pascal_case(input_string: str):
     return ''.join([i.title() for i in split_string])
 
 
-def render_templates(job_types, security_environment, api_name):
+def get_states_for_jobs(job_types: dict) -> dict:
+    states = {}
```
AndrewPlayer3, line 16: Now that there is significantly more functionality here, it's probably worth some basic docstrings. If you don't feel like that can be done quickly enough, don't worry about it. I'd rather get to testing.

jtherrmann: Might just be rewritten with CDK at some point, so not going to bother with documentation.

apps/render_cf.py
jtherrmann: I guess I was trying to leave open the possibility that this function becomes like a definitive reference for our job spec syntax. But I'm fine with deleting the comment if you think it just causes confusion.

AndrewPlayer3 (approved):

PR #2453, opened 2024-10-07
Per-Task Compute Envs and Compute Env Files

AndrewPlayer3:
Compute environments are now specified at the task level. Compute environments that are shared between multiple tasks / job_specs can be defined in `config/compute_environments.yml` like:
```yaml
compute_environments:
  ComputeEnvName1:
    instance_types: foo
  ComputeEnvName2:
    instance_types: bar
```
Then, they can be called in the tasks like:
```yaml
tasks:
  ...
  compute_environment:
    import: 'ComputeEnvName1'
  ...
```
This method allows us to avoid repeated definitions of compute environments, while still allowing for the information to be spread across as few places as possible. An alternative would be to place the compute envs each in their own files (i.e. `config/[job_spec_name].yaml`), but that is more complicated to implement and spreads a small amount of information over more files.


jtherrmann:
Just to clarify, compute environments can no longer be defined at the top level of the job spec, right? They either have to be defined at the task level *or* in `COMPUTE_ENVIRONMENTS.yml`? That's my preferred behavior, just wanted to confirm that's how it's implemented.

job_spec/SRG_GSLC.yml
```diff
@@ -61,6 +56,8 @@ SRG_GSLC:
         - Ref::bucket_prefix
         - Ref::granules
       timeout: 10800
+      compute_environment:
+        from_file: SrgGslc
```
jtherrmann, line 60: I might prefer something like `import` instead of `from_file`, just because when I first read `from_file: 'ComputeEnvName1'`, I thought it meant the compute environment was defined in a file named `ComputeEnvName1`.

job_spec/compute_environments/COMPUTE_ENVIRONMENTS.yml
jtherrmann, line 1: I'm fine with having a single file for compute environments, but I think I'd prefer it go at the top level of the repo, and maybe use lowercase, so `compute_environments.yml`. It seems odd to have a `compute_environments/` subdirectory for a single file.

AndrewPlayer3:
> Just to clarify, compute environments can no longer be defined at the top level of the job spec, right? They either have to be defined at the task level _or_ in `COMPUTE_ENVIRONMENTS.yml`? That's my preferred behavior, just wanted to confirm that's how it's implemented.

That is correct.

apps/render_cf.py
```diff
@@ -92,8 +92,13 @@ def get_job_parameters(item: str, items: str, job_spec: dict) -> dict:
...
 
-def get_step_for_batch_submit_job(task: dict, job_spec: dict) -> dict:
-    compute_environment = job_spec['compute_environment']['name']
+def get_step_for_batch_submit_job(task: dict) -> dict:
+    compute_environment = None
+    if 'compute_environment' in task:
+        if 'import' in task['compute_environment']:
+            compute_environment = task['compute_environment']['import']
+        else:
+            compute_environment = task['compute_environment']['name']
```
jtherrmann, line 101: We're requiring that each task always includes a `compute_environment`, right? If yes, then we could just simplify like this and let `KeyError` be raised if task doesn't include `compute_environment`:

```suggestion
    if 'import' in task['compute_environment']:
        compute_environment = task['compute_environment']['import']
    else:
        compute_environment = task['compute_environment']['name']
```

apps/render_cf.py
```diff
@@ -185,6 +192,7 @@ def render_costs(job_types: dict, cost_profile: str) -> None:
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument('-j', '--job-spec-files', required=True, nargs='+', type=Path)
+    parser.add_argument('-e', '--compute-environment-files', nargs='+', type=Path)
```
jtherrmann, line 195: Let's at least simplify the `--compute-environment-files` option so that it takes a single string (and rename it to `--compute-environment-file`), to make it clear that there's only one compute environment file.

apps/compute-cf.yml.j2
```diff
@@ -1,5 +1,26 @@
 AWSTemplateFormatVersion: 2010-09-09
 
+
+{%- set compute_envs = [] -%}
+{%- set compute_env_names = [] -%}
+{%- for job_type, job_spec in job_types.items() -%}
+  {%- for task in job_spec['tasks'] -%}
+    {% set env = task['compute_environment'] %}
+    {% if 'name' in env and env['name'] != 'Default' and env['name'] not in compute_env_names %}
```
jtherrmann, line 9: Does the `env['name'] not in compute_env_names` condition allow multiple tasks to define compute environments with the same name, and uses whichever one is defined first, and then the remaining compute environments get silently ignored? If that's the current behavior, then I'd prefer to instead raise an exception if multiple compute environments have the same name.

apps/compute-cf.yml.j2
```diff
@@ -1,5 +1,26 @@
...
+      {%- do compute_envs.append(env) -%}
+      {%- do compute_env_names.append(env['name']) -%}
+    {% endif %}
+    {% if 'import' in env and env['import'] not in compute_env_names %}
+      {% set name = env['import'] %}
+      {% set env = compute_env_imports[name] %}
+      {% do env.update({'name': name}) %}
```
jtherrmann, line 16: Does this work in Jinja?
```suggestion
      {% do env['name'] = name %}
```

jtherrmann (commented):
I'd be curious to see what it would look like to construct the list of compute environments in Python and then pass in the list to Jinja, to reduce the amount of Jinja we're adding in this PR.

apps/render_cf.py
```diff
@@ -158,6 +164,45 @@ def render_templates(job_types, security_environment, api_name):
...
+            if 'name' in compute_env:
+                name = compute_env['name']
+                if name in compute_env_names:
+                    raise NameError(
+                        f'Compute envs must have unique names but the following is defined more than once: {name}.'
+                    )
+                compute_envs.append(compute_env)
+                compute_env_names.append(name)
+            elif 'import' in compute_env and compute_env['import'] != 'Default':
+                compute_env_imports.add(compute_env['import'])
```
jtherrmann, line 183: Should probably have a final `else` (feel free to replace the `assert` with a `ValueError`):
```suggestion
            elif 'import' in compute_env and compute_env['import'] != 'Default':
                compute_env_imports.add(compute_env['import'])
            else:
                assert compute_env['import'] == 'Default'
```

apps/render_cf.py
```diff
@@ -158,6 +164,45 @@ def render_templates(job_types, security_environment, api_name):
...
+    compute_env_imports = set()
+    for _, job_spec in job_types.items():
+        for task in job_spec['tasks']:
+            compute_env = task['compute_environment']
+            if 'name' in compute_env:
+                name = compute_env['name']
+                if name in compute_env_names:
+                    raise NameError(
+                        f'Compute envs must have unique names but the following is defined more than once: {name}.'
+                    )
```
jtherrmann, line 179: My understanding is that the builtin `NameError` exception is typically raised when an undefined variable is referenced, so I don't think it's intended to be raised in other contexts. I think `ValueError` is most appropriate here, thought we could probably replace the entire `if name in compute_env_names` block with `assert name in compute_env_names` and that would be fine too.
```suggestion
                    raise ValueError(
                        f'Compute envs must have unique names but the following is defined more than once: {name}.'
                    )
```

apps/render_cf.py
```diff
@@ -158,6 +164,45 @@ def render_templates(job_types, security_environment, api_name):
         template_file.with_suffix('').write_text(output)
 
 
+def get_compute_environments(job_types: dict, compute_env_file: Path) -> list[dict]:
```
jtherrmann, line 167: Since `compute_env_file` can be `None`:
```suggestion
def get_compute_environments(job_types: dict, compute_env_file: Optional[Path]) -> list[dict]:
```

apps/render_cf.py
```diff
@@ -158,6 +164,45 @@ def render_templates(job_types, security_environment, api_name):
...
+                compute_envs.append(compute_env)
+                compute_env_names.append(name)
+            elif 'import' in compute_env and compute_env['import'] != 'Default':
+                compute_env_imports.add(compute_env['import'])
+
+    if compute_env_file:
+        compute_envs_from_file = yaml.safe_load(compute_env_file.read_text())['compute_environments']
+        for name in compute_envs_from_file:
+            if name in compute_env_names:
+                raise NameError(
```
jtherrmann, line 189: See previous comment.
```suggestion
                raise ValueError(
```

apps/render_cf.py
```diff
@@ -158,6 +164,45 @@ def render_templates(job_types, security_environment, api_name):
...
+            compute_env = compute_envs_from_file[name]
+            compute_env['name'] = name
+            compute_envs.append(compute_env)
+            compute_env_names.append(name)
+
+    for name in compute_env_imports:
+        if name not in compute_envs_from_file:
+            raise NotImplementedError(
+                f'The following compute env is imported but not defined in the compute envs file: {name}.'
+            )
```
jtherrmann, line 201: I think there's a stronger argument for using `NotImplementedError` here than there is for using `NameError` in the examples above, but the docstring for `NotImplementedError` says `Method or function hasn't been implemented yet`, so I feel like it's a builtin exception that's only intended to be raised when you call a function or method that has been defined but hasn't yet been implemented.

In general, I think builtin exceptions have pretty specific intended use cases and it's usually better to define a custom exception when `ValueError` isn't specific enough (or just use `assert` for non-user-facing code).
```suggestion
            raise ValueError(
                f'The following compute env is imported but not defined in the compute envs file: {name}.'
            )
```

apps/render_cf.py
```diff
@@ -158,6 +164,45 @@ def render_templates(job_types, security_environment, api_name):
         template_file.with_suffix('').write_text(output)
 
 
+def get_compute_environments(job_types: dict, compute_env_file: Path) -> list[dict]:
+    compute_envs = []
+    compute_env_names = []
```
jtherrmann, line 169: Could make `compute_env_names` a set to make it more clear what it's being used for, which is just to check if it contains a given name. When you need to check if an item exists in a collection, my go-to is always `set`, for both efficiency and readability.
```suggestion
    compute_env_names = set()
```

apps/compute-cf.yml.j2
```diff
@@ -29,8 +29,8 @@ Outputs:
   JobQueueArn:
     Value: !Ref BatchJobQueue
 
-  {% for job_type, job_spec in job_types.items()  if 'Default' != job_spec['compute_environment']['name'] %}
-  {% set name = job_spec['compute_environment']['name'] %}
+  {% for env in compute_envs %}
+  {% set name = env['name'] %}
```
jtherrmann, line 33:
```suggestion
  {% for name in compute_env_names %}
```

apps/render_cf.py
jtherrmann: We could also set it to `{'Default'}` so that an error will be raised if someone tries to define a compute environment named `Default`?

apps/compute-cf.yml.j2
jtherrmann, line 1: Hmm, I honestly wonder if it's worth defining our default compute environment in the `compute_environments.yml` file with the name `Default` so that it's no different from our other compute environments. It would remove the special handling of the `Default` name in `render_cf.py`, and it would make more sense with the `import: Default` syntax. I know we dismissed this idea because it would be extra work, but it seems like we have all the pieces in place to do it pretty easily, and it would simplify our CloudFormation and rendering code. What do you think?

jtherrmann: I think you could just create a `Default` compute environment definition in `compute_environments.yml` and then delete the `ComputeEnvironment` and `BatchJobQueue` resources in `compute-cf.yml.j2`. This would mean those resources get deleted and then re-created with different names when we do our deployments, but I'm not actually sure that would be a problem.

apps/render_cf.py
jtherrmann: Disregard the `{'Default'}` suggestion if we end up defining `Default` with our other compute environments.

apps/render_cf.py
```diff
@@ -158,6 +165,47 @@ def render_templates(job_types, security_environment, api_name):
         template_file.with_suffix('').write_text(output)
 
 
+def get_compute_environments(job_types: dict, compute_env_file: Optional[Path]) -> list[dict]:
+    compute_envs = []
+    compute_env_names = set()
```
jtherrmann, line 170: How do you feel about this as a way to ensure that an error is raised if someone tries to define a `Default` compute environment?
```suggestion
    compute_env_names = {'Default'}
```

jtherrmann: Meh, I think that's a bit confusing since we later construct a `compute_env_names` variable which doesn't include `'Default'`. Maybe let's add an explicit check for this below.

jtherrmann (approved):

PR #2454, opened 2024-10-10
Create default compute env

AndrewPlayer3:


jtherrmann:
I had asked @AndrewPlayer3 to try moving the default compute environment definition into `job_spec/config/compute_environments.yml` as part of the work we're doing for <https://asfdaac.atlassian.net/browse/TOOL-2860>, but I realized after seeing the implementation that this would remove the flexibility of being able to specify certain compute environment parameters (`instance_types` and `ami_id`) at the deployment level OR the job spec level, and would force us to only define those at the job spec level. So, unfortunately, I don't think we would want to merge these changes at this time.

PR #2463, opened 2024-10-18
Release v8.0.0

jtherrmann:
Before merging:

- [x] https://github.com/ASFHyP3/hyp3-srg/pull/57
- [x] Drain job queues for prod deployments
- [x] Revert changes to the `hyp3-multi-burst-sandbox` deployment (it may still be useful for multi-burst development)
- [x] Validate all changes
- [x] Regression testing
- [x] Fix the compute envs bug described below.


apps/render_cf.py
```diff
@@ -11,35 +12,245 @@ def snake_to_pascal_case(input_string: str):
...
+        for name in compute_envs_from_file:
+            assert name != 'Default'
+            if name in compute_env_names:
+                raise ValueError(
+                    f'Compute envs must have unique names but the following is defined more than once: {name}.'
+                )
+            compute_env = compute_envs_from_file[name]
+            compute_env['name'] = name
+            compute_envs.append(compute_env)
+            compute_env_names.add(name)
```
jtherrmann, line 227: This results in all of the compute envs defined in `job_spec/config/compute_environments.yml` getting added to the `compute_envs` list, so they get created in all deployments, even if those compute envs were not actually imported in any of the job specs in the `job_types` list.

jtherrmann: We may also want to add unit tests for this function (and probably some of the other render functions too).

AndrewPlayer3 (dismissed):

AndrewPlayer3 (approved):

PR #2465, opened 2024-10-18
Compute Envs Must be Imported to Render

AndrewPlayer3:


job_spec/config/compute_environments.yml
```diff
@@ -1,9 +1,33 @@
+# Compute Environments that are shared between multiple jobs and/or steps should be defined here,
```
AndrewPlayer3, line 1: This is a bit out-of-place here, but I think its valuable. We can remove it whenever we properly document job spec syntax.

Makefile
```diff
@@ -1,4 +1,5 @@
 API = ${PWD}/apps/api/src
+RENDER_CF = ${PWD}/apps
```
jtherrmann, line 2:
```suggestion
APPS = ${PWD}/apps
```

Makefile
```diff
@@ -10,7 +11,7 @@ DISABLE_PRIVATE_DNS = ${PWD}/apps/disable-private-dns/src
 UPDATE_DB = ${PWD}/apps/update-db/src
 UPLOAD_LOG = ${PWD}/apps/upload-log/src
 DYNAMO = ${PWD}/lib/dynamo
-export PYTHONPATH = ${API}:${CHECK_PROCESSING_TIME}:${GET_FILES}:${HANDLE_BATCH_EVENT}:${SET_BATCH_OVERRIDES}:${SCALE_CLUSTER}:${START_EXECUTION_MANAGER}:${START_EXECUTION_WORKER}:${DISABLE_PRIVATE_DNS}:${UPDATE_DB}:${UPLOAD_LOG}:${DYNAMO}
+export PYTHONPATH = ${API}:${CHECK_PROCESSING_TIME}:${GET_FILES}:${HANDLE_BATCH_EVENT}:${SET_BATCH_OVERRIDES}:${SCALE_CLUSTER}:${START_EXECUTION_MANAGER}:${START_EXECUTION_WORKER}:${DISABLE_PRIVATE_DNS}:${UPDATE_DB}:${UPLOAD_LOG}:${DYNAMO}:${RENDER_CF}
```
jtherrmann, line 14:
```suggestion
export PYTHONPATH = ${API}:${CHECK_PROCESSING_TIME}:${GET_FILES}:${HANDLE_BATCH_EVENT}:${SET_BATCH_OVERRIDES}:${SCALE_CLUSTER}:${START_EXECUTION_MANAGER}:${START_EXECUTION_WORKER}:${DISABLE_PRIVATE_DNS}:${UPDATE_DB}:${UPLOAD_LOG}:${DYNAMO}:${APPS}
```

jtherrmann (commented):
Suggest changing to `APPS` since it's not specific to `render_cf.py`

apps/render_cf.py
```diff
@@ -192,45 +192,63 @@ def render_templates(job_types, compute_envs, security_environment, api_name):
...
+            raise ValueError(
+                f'The following compute env is imported but not defined in the compute envs file: {name}.'
+            )
+
+    return compute_envs
+
+
+def get_compute_environments(job_types: dict, compute_env_file: Optional[Path] = None) -> list[dict]:
+    compute_envs = []
+    compute_env_names = set({'Default'})
```
jtherrmann, line 224:
```suggestion
    compute_env_names = {'Default'}
```

apps/render_cf.py
```diff
@@ -192,45 +192,63 @@ def render_templates(job_types, compute_envs, security_environment, api_name):
...
+
+    for name in compute_envs_from_file:
+        if name in compute_env_imports:
+            if name in compute_env_names:
+                raise ValueError(
+                    f'Compute envs must have unique names but the following is defined more than once: {name}.'
+                )
+            compute_envs_from_file[name].update({'name': name})
+            compute_envs.append(compute_envs_from_file[name])
+            compute_env_names.add(name)
```
jtherrmann, line 211: Should it be more like this? So that the error gets raised for duplicate names even if they're not imported?
```suggestion
    for name in compute_envs_from_file:
        if name in compute_env_names:
            raise ValueError(
                f'Compute envs must have unique names but the following is defined more than once: {name}.'
            )
        if name in compute_env_imports:
            compute_envs_from_file[name].update({'name': name})
            compute_envs.append(compute_envs_from_file[name])
        compute_env_names.add(name)
```

job_spec/config/compute_environments.yml
```diff
@@ -1,9 +1,33 @@
...
+#   steps:
+#     step_name:
+#       ...
+#       compute_environment:
+#         import: 'ExampleComputeEnvName'
+
+# Note that the name must be unique, i.e. a compute environment that is defined in a job spec using
+# the `name` key cannot also be defined here, unless it is not imported in the same deployment (envs here
+# are only rendered if they are imported). Also note that the name should be in pascal case to avoid
+# cfn-lint errors.
```
jtherrmann, line 24: This paragraph was somewhat difficult for me to parse -- is it just describing error cases that will be caught automatically? i.e. duplicate names should be caught by `render_cf.py`, and I assume non-pascal-case names would also lead to an error? If so, I suggest we just remove it and let those error messages speak for themselves.
```suggestion
```

job_spec/config/compute_environments.yml
jtherrmann: I like it for now. Yes, as I was writing the Changelog entry for https://github.com/ASFHyP3/hyp3/pull/2463 I was thinking we just need a formal specification of our job spec language.

apps/render_cf.py
AndrewPlayer3: I did it like that intentionally so that they could be overridden by different deployments, which seems in line with how a lot of other things work. I feel like there was another reason I did that too but I can't remember.

jtherrmann: Hm OK, let me read it more carefully

jtherrmann: Yep, I think that makes sense.

AndrewPlayer3: Honestly, I'll probably change it to your suggestion. I don't really love that behavior the more I think about it. I'm just playing around with it to see if I remember what the other issue was.

job_spec/config/compute_environments.yml
AndrewPlayer3: I mostly included that to highlight the 'unless it is not imported in the same deployment' part. I'll remove that if I remove the behavior.

jtherrmann (approved):

PR #2501, opened 2024-11-22
Add job spec for opera-disp-tms

forrestfwilliams:


job_spec/OPERA_DISP_TMS.yml
```diff
@@ -0,0 +1,79 @@
...
+          description: End of secondary date range to generate a tile map for (%Y%m%d)
+          type: string
+          minLength: 8
+          maxLength: 8
+          example: 20200814
+  cost_profiles:
+    DEFAULT:
+      cost: 1.0
+  validators:
+    - check_bounds_formatting
```
jtherrmann, line 59: We had added the [`check_bounds_formatting`](https://github.com/ASFHyP3/hyp3/blob/develop/apps/api/src/hyp3_api/validation.py#L143) validator for use by [`SRG_TIME_SERIES`](https://github.com/ASFHyP3/hyp3/blob/develop/job_spec/SRG_TIME_SERIES.yml). It looks like the `minlon minlat maxlon maxlat` ordering is compatible with the validator, but keep in mind the validator also allows for passing `0.0, 0.0, 0.0, 0.0` bounds, which has a special meaning for `SRG_TIME_SERIES` but may not be applicable for this job type, but perhaps we don't mind allowing `0.0, 0.0, 0.0, 0.0` bounds for this job type anyway?

job_spec/OPERA_DISP_TMS.yml
```diff
@@ -0,0 +1,79 @@
...
+          minLength: 8
+          maxLength: 8
+          example: 20200101
+      end_date:
+        api_schema:
+          description: End of secondary date range to generate a tile map for (%Y%m%d)
+          type: string
+          minLength: 8
+          maxLength: 8
+          example: 20200814
```
jtherrmann, line 54: There's an extra level of indentation here:
```suggestion
  parameters:
    tile_type:
      api_schema:
        description: Type of tile map to create
        type: string
        minLength: 8
        maxLength: 15
        enum:
          - displacement
          - secant_velocity
          - velocity
        example: displacement
    bounds:
      api_schema:
        description: Integer bounds in format (minlon minlat maxlon maxlat)
        type: array
        minItems: 4
        maxItems: 4
        example:
          - -121
          - 38
          - -120
          - 39
    orbit_direction:
      api_schema:
        description: Orbit direction of data
        type: string
        minLength: 9
        maxLength: 10
        enum:
          - ascending
          - descending
        example: descending
    start_date:
      api_schema:
        description: Start of secondary date range to generate a tile map for (%Y%m%d)
        type: string
        minLength: 8
        maxLength: 8
        example: 20200101
    end_date:
      api_schema:
        description: End of secondary date range to generate a tile map for (%Y%m%d)
        type: string
        minLength: 8
        maxLength: 8
        example: 20200814
```

job_spec/OPERA_DISP_TMS.yml
forrestfwilliams: Yeah that's an edge case I'm OK with for this job.

job_spec/OPERA_DISP_TMS.yml
forrestfwilliams: Thanks for the suggestions Jake!

PR #2566, opened 2025-01-22
Containerized deployment environment

### LINKED ISSUE #2551: Verify Python==3.13 when running `make build`
Per the README, when manually building/deploying HyP3, the user needs to make sure they're running Python v3.13 when pip installing the lambda dependencies:

> - Install Python dependencies for AWS Lambda functions (requires pip for python 3.13)
> ```
> make build
> ```

Missing this step and building with a different version of python can lead to import or runtime errors when the lambda functions run after being deployed.

It should be simple to assert `python --version` starts with `3.13` within the `make build` target itself, to prevent this error up front in the future.

https://github.com/ASFHyP3/hyp3/blob/develop/Makefile#L17

Alex-Lewandowski:
This PR is intended to simplify the HyP3 deployment process by organizing build parameters into a `.env` and providing a Docker image configured to deploy Lambdas with Python3.13 runtimes via CloudFormation. 

Closes #2551 

There is nothing included in this PR that prevents deploying HyP3 in the way maintainers always have, but it adds opt-in convenience. An example use-case is deploying HyP3 from a Mac, where iOS-specific Python package binaries may be incompatible with the AWS Lambda runtime.

- Instructions updated in `README.md`
- `hyp3.example` is a template used to create a deployment-specific `hyp3.env` file containing all build parameters as environment variables. Some vars are required and some are optional. CloudFormation will use default parameters for optional env vars that are commented or have empty values in `hyp3.env`. `hyp3.env` has been added to `.gitignore` to help prevent pushing semi-sensitive info to a public repo.
- `Dockerfile` defines an image built off of `amazon/aws-lambda-python:3.13`, installs make and awscli, and copies in the hyp3 dir
- `Makefile` updates:
  - All targets marked PHONY (since none produce output named for the target)
  - New make targets:
    - `image`: Creates an image from `Dockerfile`
    - `shell`: Runs a /bin/bash shell on the image, mounts the hyp3 dir, and pulls in user's AWS creds for the AWS_DEFAULT_PROFILE, as defined in `hyp3.env`, or defaults to the `default` profile. Assumes the user has AWS configured with a profile matching their provided AWS_DEFAULT_PROFILE.
    - `package`: Packages the CloudFormation template
    - `deploy`: Deploys HyP3 with CloudFormation
    - `help`: Lists all targets and their descriptions. Provides instructions for deploying with make
    - `clean`: Added removal of the deployment image if it is present, Docker is installed, and Docker Engine is running

To deploy with a deployment image:
1. Configure a `hyp3.env` 
2. `make image`
3. `make shell`
4. `make install`
5. `make build`
6. `make package`
7. `make deploy`

To do it the old way:
1. Set up your local deployment environment
2. `make install`
3. `make build`
4. Manually run `aws cloudformation package ...`
5. Manually run `aws cloudformation deploy ...`


hyp3.example
jhkennedy, line 1: In the .gitignore, this is named `hyp3.env.example` so we should probably rename the file.

I'll have to think about where this should live a little, but I'm inclined to put it in the docs directory if it's an example...

README.md
```diff
@@ -61,52 +63,55 @@ Review the parameters in [cloudformation.yml](apps/main-cf.yml) for deploy time
 
 ### Deploy with CloudFormation
```
jhkennedy, line 65: It might be worth a note/admonishion here to say that manual deployments should only be done to temporary deploymetns, otherwise, they should be hooked into our CI/CD infrastructure... 

I'll think on that a little.

Makefile
```diff
@@ -25,39 +86,120 @@ build: render
...
+.PHONY: package
+package: ## Run 'make install && make build', and package CloudFormation artifacts
+	aws cloudformation package \
+		--template-file ./apps/main-cf.yml \
+		--s3-bucket "$$ARTIFACT_BUCKET_NAME" \
+		--output-template-file packaged.yml; \
+
+
+.PHONY: deploy
+deploy: ## Deploy HyP3 with AWS CloudFormation
```
jhkennedy, line 171: I think these might deprecate [.github/actions/deploy-hyp3](https://github.com/ASFHyP3/hyp3/tree/develop/.github/actions/deploy-hyp3). Though that may require writing a `.env` file as part of deployment...

jhkennedy (commented):
Overall, I like this! I haven't done an extensive review, but did jot down some thoughts for discussion.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN mkdir /hyp3
+WORKDIR /hyp3
+
+COPY . /hyp3/
+
+RUN python3 -m pip install --no-cache-dir --upgrade wheel && \
+    python3 -m pip install --no-cache-dir awscli
+
+ENTRYPOINT ["/bin/bash"]
```
jhkennedy, line 17:
```suggestion
ENTRYPOINT ["/bin/bash"]

```

Makefile
```diff
@@ -11,10 +13,69 @@ DISABLE_PRIVATE_DNS = ${PWD}/apps/disable-private-dns/src
...
+DOCKER_RUN = docker run --rm -it \
+	--env-file $(ENV_FILE) \
+	-v ~/.aws/:/root/.aws/:ro \
+	-v /tmp/aws-cli-cache:/root/.aws/cli/cache \
+	-v ${PWD}:/hyp3/ \
+	-e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY \
+	-e AWS_DEFAULT_PROFILE \
+	-e AWS_DEFAULT_REGION \
+	-e AWS_DEFAULT_ACCOUNT \
+	${DEPLOY_ENV_IMAGE_NAME}:latest
```
jhkennedy, line 41:
```suggestion
	${DEPLOY_ENV_IMAGE_NAME}:latest
```
I'm not sure we'll want to use the latest tag here, or a specific version number...

Makefile
```diff
@@ -25,39 +86,120 @@ build: render
...
+			[REQUIRED_SURPLUS]="RequiredSurplus" \
+			[SECRET_ARN]="SecretArn" \
+			[SUBNET_IDS]="SubnetIds" \
+			[SYSTEM_AVAILABLE]="SystemAvailable" \
+			[VPC_ID]="VpcId" \
+		); \
+		for var in "$${!PARAMS[@]}"; do \
+			value="$${!var}"; \
+			[ -n "$$value" ] && PARAM_OVERRIDES="$$PARAM_OVERRIDES $${PARAMS[$$var]}=$$value"; \
+		done; \
```
jhkennedy, line 198: This seems a little overly complex for me compared to:
https://github.com/ASFHyP3/hyp3/blob/develop/.github/actions/deploy-hyp3/action.yml#L112-L135

And looks like it doesn't handle these special cases:
https://github.com/ASFHyP3/hyp3/blob/develop/.github/actions/deploy-hyp3/action.yml#L93-L106

hyp3.example
Alex-Lewandowski: Good catch!

Makefile
Alex-Lewandowski: As this is currently organized, for the manual deployment workflow, build parameters are provided in `hyp3.env`, and not all are required. They are loaded as env vars when running the image. Any unused optional parameters end up as env vars with empty string values. This bit of code in the `deploy` target only adds parameter overrides for env vars with valid string values. The unused optional parameters are not included in the overrides and so defaults from the CF template will be used instead. 

In the CICD deployment workflow, it looks like all the parameters are defined in `deploy-*.yml`, some of which are retrieved from GitHub secrets. 

So, for CICD, by the time the deployment action is run, all the needed parameter overrides are known so you can hardcode which to include in the CF deploy command. When deploying from a local environment, I have to check which parameters were actually set in `hyp3.env` and only include those in the overrides. Otherwise, I would provide some of the parameter overrides values as empty strings, breaking the deployment.

jtherrmann:
My only concern with this PR is that it adds another method of deploying HyP3, when I'd prefer to see us standardize around one approach, preferably sticking to an infrastructure-as-code approach.

I'm not a fan of the [deployment section in our README](https://github.com/ASFHyP3/hyp3#deployment) because it recommends deploying from the command-line, while most of our actual deployments consist of a GitHub environment + GitHub Actions deployment workflow. For example, our https://hyp3-opera-disp-sandbox.asf.alaska.edu deployment consists of a [`hyp3-opera-disp-sandbox`](https://github.com/ASFHyP3/hyp3/settings/environments/4679251322/edit) GitHub environment and a [`deploy-opera-disp-sandbox.yml`](https://github.com/ASFHyP3/hyp3/blob/develop/.github/workflows/deploy-opera-disp-sandbox.yml) deployment workflow.

I typically adapt the steps from our [New HyP3 Enterprise deployment](https://github.com/ASFHyP3/.github-private/wiki/New-HyP3-Enterprise-deployment) wiki article whenever I set up a new deployment (regardless of whether it's an enterprise deployment), but I'd propose moving some version of those steps into the HyP3 README.

I know that @asjohnston-asf and maybe others have favored deploying from the command-line as shown in our README, which is understandable since it's a quick way of getting a deployment up and running, but I'd prefer that not be our standard recommended approach, since I'd rather all long-lived HyP3 deployments be managed via git / GitHub.

But that's just my opinion, I don't think we ever reached consensus on this within the HyP3 team.

jhkennedy:
> My only concern with this PR is that it adds another method of deploying HyP3, when I'd prefer to see us standardize around one approach, preferably sticking to an infrastructure-as-code approach.

I agree with this, and we shouldn't do manual deployments. Maybe even more so now that multiple teams will be working on it. 

In particular, HyP3 currently has an unwritten contract that all deployments are always up to date, so Vertex and the SDK only have to work with the latest version of HyP3. Keeping everything in IaC facilitates that and following the same processes ensures that every HyP3 deployment is listed here:
https://github.com/ASFHyP3/hyp3/settings/environments

If we start changing that contract, we're going to need to be a whole lot better about reporting version numbers and checking for compatibility. 

> I'm not a fan of the [deployment section in our README](https://github.com/ASFHyP3/hyp3#deployment) ... 
> I typically adapt the steps from our [New HyP3 Enterprise deployment](https://github.com/ASFHyP3/.github-private/wiki/New-HyP3-Enterprise-deployment) wiki

Same on the README! We should document how we want HyP3 to be deployed there, so the wiki article should be moved largely to the README. Unfortunately, that wiki is private, so @Alex-Lewandowski couldn't have followed it even if he wanted to since he's not in this org yet. 

I expect there are a ton of places where Tools shared knowledge and SoPs will need to be written down and exposed for multi-team development of HyP3.

README.md
jhkennedy: more thoughts around this in this comment:
https://github.com/ASFHyP3/hyp3/pull/2566#issuecomment-2613451411

jhkennedy:
All that said, I do like: using docker to build out deployment packages. This potentially avoids some hard-to-resolve edge cases, especially around linking.

I do, however,  like potentially dropping the custom github action so all of the "what you do with this repo" stuff is described in the makefile, but  I don't really want to support `.env` files for the environment variables to discourage non IaC deployments.

Alex-Lewandowski:
>I'd prefer to see us standardize around one approach, preferably sticking to an infrastructure-as-code approach.

I agree with this 100%. I much prefer standardizing around an IoC approach. I took a naive pass at deploying HyP3 based on the README and tried to add a little automation to that process, but there is no need to reinvent a squarer, less performant wheel.

My goal is to walk through the process of deploying HyP3 following the standardized process and pass on what I learn to the rest of the Services team. Can I have access to the [HyP3 Enterprise deployment wiki](https://github.com/ASFHyP3/.github-private/wiki/New-HyP3-Enterprise-deployment)?

It sounds like you already have a documented IoC approach with GitHub actions in place so most of this PR can go away. We could also convert it to a draft and use it for a README update based on the wiki. I'd love to step through it and help identify any areas that might be confusing to someone new to the HyP3 backend.

jhkennedy:
> Can I have access to the [HyP3 Enterprise deployment wiki](https://github.com/ASFHyP3/.github-private/wiki/New-HyP3-Enterprise-deployment)?

@Alex-Lewandowski , yep, working on it!

PR #2600, opened 2025-02-14
Adding test environment for Services

Cameronsplaze:


.github/workflows/deploy-enterprise-test.yml
```diff
@@ -97,6 +97,26 @@ jobs:
             ami_id: /aws/service/ecs/optimized-ami/amazon-linux-2023/recommended/image_id
             distribution_url: ''
 
+          - environment: hyp3-osl-test
+            domain: hyp3-osl-test.asf.alaska.edu
+            template_bucket: services-hyp3-artifacts
+            image_tag: latest
```
jtherrmann, line 103: You might consider using the `test` image tag, since that will allow you to use the latest `develop` branch builds of https://github.com/ASFHyP3/hyp3-autorift/ and <https://github.com/ASFHyP3/hyp3-isce2/>. But if you want to only use the versions that have been released (merged to `main`), then you'll want to stick with the `latest` tag. (All of the other deployments in this file use the `test` tag, but I don't necessarily see that inconsistency as a problem.)
```suggestion
            image_tag: test
```

jtherrmann: Also note that the deployments in this file only get triggered on a merge to hyp3's `develop` branch, so if you want to make frequent changes to this deployment without having to make a PR every time, you might consider setting up a sandbox deployment that deploys from a separate feature branch, similar to [hyp3-multi-burst-sandbox](https://github.com/ASFHyP3/hyp3/blob/develop/.github/workflows/deploy-multi-burst-sandbox.yml), which deploys from the `multi-burst-sandbox` branch.

This would allow the OSL team to use this deployment for rapid experimentation and development. For example, you might want to change the job types available in the `job_files` field, or make changes to the job specs themselves (e.g. to change the API interface or specify a custom image tag that corresponds to a particular build of a plugin that you're working on).

Do you foresee changing which job types are available in this deployment, or using this deployment for HyP3 plugin development?

PR #2602, opened 2025-02-15
publish aria_s1_gunw jobs for ingest via ingest-adapter

asjohnston-asf:
https://github.com/ASFHyP3/ingest-adapter/

### DIFF
--- a/job_spec/ARIA_S1_GUNW.yml
+++ b/job_spec/ARIA_S1_GUNW.yml
@@ -59,7 +59,7 @@ ARIA_S1_GUNW:
         - --bucket
         - '!Ref Bucket'
         - --bucket-prefix
-        - Ref::bucket_prefix
+        - Ref::job_id
         - --reference-scenes
         - Ref::reference
         - --secondary-scenes
@@ -83,7 +83,7 @@ ARIA_S1_GUNW:
         - --bucket
         - '!Ref Bucket'
         - --bucket-prefix
-        - Ref::bucket_prefix
+        - Ref::job_id
         - --weather-model
         - HRRR
       timeout: 10800  # 3 hr
--- a/job_spec/ARIA_S1_GUNW.yml
+++ b/job_spec/ARIA_S1_GUNW.yml
@@ -95,3 +95,18 @@ ARIA_S1_GUNW:
         - EARTHDATA_PASSWORD
         - ESA_USERNAME
         - ESA_PASSWORD
+    - name: PUBLISH
+      image: ghcr.io/ASFHyP3/ingest-adapter
+      command:
+        - --source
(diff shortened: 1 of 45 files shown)


job_spec/ARIA_S1_GUNW.yml
```diff
@@ -95,3 +95,19 @@ ARIA_S1_GUNW:
...
         - ESA_PASSWORD
+    - name: PUBLISH
+      image: ghcr.io/ASFHyP3/ingest-adapter
+      command:
+        - --source
+        - https://hyp3-test-api.asf.alaska.edu
+        - --job-id
+        - Ref::bucket_prefix
+        - --target
+        - UAT
```
asjohnston-asf, line 106: Specifying the job ID to publish is easy, we already have that immediately available as `bucket_prefix`.

Specifying "from this hyp3 environment to this CMR environment" is trickier. We wouldn't want to hardcode them right here, because we want jobs running in hyp3-test-api to be published to cmr uat, and jobs running in hyp3-api to be published to cmr prod.

We could hardcode the business logic in the ingest-adapter container to decide which hyp3 environments publish to which cmr environment. The `secrets` block might be a better place to provide those inputs, since that lets us use the same job spec in every hyp3 environment and just override the secrets in each environment.

asjohnston-asf: Maybe:

```
command:
  - --job-id
  - Ref::bucket_prefix
secrets:
  - EARTHDATA_USERNAME
  - EARTHDATA_PASSWORD
  - INGEST_SNS_TOPIC_ARN
  - HYP3_URL
```

That feels a little clunky, since the container would have to assume those inputs come via environment variables rather than the command line, but it's a quick and easy way to provide hyp3-deployment-specific parameters without having to add them as hyp3 stack parameters and plumb them down into the job spec.

asjohnston-asf: I realized this call won't need to include EDL credentials. All the publish container will do is send a message including the hyp3 url and job id to the ingest adapter sns topic; the ingest adapter app deployed behind that topic will handle the hyp3/cmr queries. We'll have two adapter deployments, one integrated with CMR UAT and one integrated with CMR prod. The publish container just needs to point at one SNS topic or the other depending on which CMR environment the particular hyp3 deployment should publish to.

jtherrmann:
Confirmed that it matches the latest README at https://github.com/ASFHyP3/ingest-adapter/pull/23

jtherrmann (approved):

jtherrmann:
Sorry I forgot about this PR when I opened https://github.com/ASFHyP3/hyp3/pull/2619 but looks like they were identical

PR #2633, opened 2025-02-28
Update entrypoint to dist-s1

cmarshak:
Used `python -um` and updated entrypoint for the docker image.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Changed
 - OPERA-DIST-S1 runtime increases from 3 to 6 hours for experimentation.
+- Updated the DIST-S1 entrypoint of the image and changed the container respectively.
 
 
 ## [9.5.1]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Changed
 - OPERA-DIST-S1 runtime increases from 3 to 6 hours for experimentation.
-- Updated the DIST-S1 entrypoint of the image and changed the container respectively.
+- Updated the DIST-S1 entrypoint of the image and changed the job spec accordingly.
 
 
 ## [9.5.1]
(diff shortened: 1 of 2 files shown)


cmarshak:
@jhkennedy

jhkennedy:
@cmarshak I'm getting this when I'm checking the docker image:
```
$ docker run ghcr.io/opera-adt/dist-s1:test --help
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "/home/ops/dist-s1/src/dist_s1/etc/entrypoint.sh": permission denied: unknown.
```

If I jump into the container, I see that `entrypoint.sh` isn't executable:
```
docker run --entrypoint /bin/bash -it --rm ghcr.io/opera-adt/dist-s1:test -l
(dist-s1-env) dist_user@92514b121c57:~$ ls /home/ops/dist-s1/src/dist_s1/etc/entrypoint.sh 
/home/ops/dist-s1/src/dist_s1/etc/entrypoint.sh
(dist-s1-env) dist_user@92514b121c57:~$ ll -lah /home/ops/dist-s1/src/dist_s1/etc/entrypoint.sh 
-rw-r--r-- 1 dist_user dist_user 59 Feb 28 05:54 /home/ops/dist-s1/src/dist_s1/etc/entrypoint.sh
```

And when I cat `entrypoint.sh` I see it trying to call a `dist_s1` CLI entrypoint, which doesn't exist in the container, but `dist-s1` does:
```
(dist-s1-env) dist_user@92514b121c57:~$ cat /home/ops/dist-s1/src/dist_s1/etc/entrypoint.sh 
#!/bin/bash --login
set -e
exec python -um dist_s1 run "$@"
(dist-s1-env) dist_user@92514b121c57:~$ which dist_s1
(dist-s1-env) dist_user@92514b121c57:~$ which dist-s1
/opt/miniforge/envs/dist-s1-env/bin/dist-s1
```

Otherwise, the parameters in the job spec do match `dist-s1 run --help`. 

--- 

As an aside, if you're using argparse, which it looks like you are, you don't need to use underscores for the arguments (e.g., `--mgrs_tile_id`) as arparse implicitly turns the dashes in arguments into underscored variable names (`--mgrs-tile-id` -> `args.mgrs_tile_id`)

cmarshak:
https://github.com/opera-adt/dist-s1/pull/48

jhkennedy (approved):

PR #2637, opened 2025-02-28
Prototype for nested cost definitions

williamh890:
- [x] https://github.com/ASFHyP3/hyp3/pull/2659
- [x] Think more about float vs. int vs. string stuff
- [x] Add multi-burst example to job spec (use hyp3-isce2 readme example)
- [x] Validate all changes in sandbox deployment, including:
  - [x] Get back expected costs when submitting jobs
  - [x] Remaining credits decrement as expected for user record
  - [x] `/costs` endpoint returns expected response


jtherrmann (commented):
Summarizing a live discussion, we decided:

- Remove the `length::` syntax and always take the length if the parameter is an array.
- Looks (`20x4`, etc.) should be the top-level key and number of bursts should be the second-level key.

lib/dynamo/dynamo/jobs.py
```diff
@@ -122,18 +122,36 @@ def _get_credit_cost(job: dict, costs: list[dict]) -> Decimal:
             if cost_definition.keys() not in ({'job_type', 'cost_parameter', 'cost_table'}, {'job_type', 'cost'}):
                 raise ValueError(f'Cost definition for job type {job_type} has invalid keys: {cost_definition.keys()}')
```
jtherrmann, line 123: I would move this check into `_get_cost_from_definition`

jtherrmann: Oh nevermind, I forgot it's a recursive function.

lib/dynamo/dynamo/jobs.py
```diff
@@ -122,18 +122,36 @@ def _get_credit_cost(job: dict, costs: list[dict]) -> Decimal:
...
+            parameter_value = len(parameter_value)
+
+        for item in cost_definition['cost_table']:
+            if item['parameter_value'] != parameter_value:
+                continue
+
+            if 'cost_table' in item:
+                return _get_cost_from_definition(job, item['cost_table'])
+            else:
+                return item['cost']
```
jtherrmann, line 150:
```suggestion
            if 'cost_table' in item:
                return _get_cost_from_definition(job, item['cost_table'])

            if item['parameter_value'] == parameter_value:
                return item['cost']
```

lib/dynamo/dynamo/jobs.py
```diff
@@ -122,18 +122,36 @@ def _get_credit_cost(job: dict, costs: list[dict]) -> Decimal:
...
-                        return item['cost']
-                raise ValueError(f'Cost not found for job type {job_type} with {cost_parameter} == {parameter_value}')
-
-            return cost_definition['cost']
+            return _get_cost_from_definition(job, cost_definition)
+
     raise ValueError(f'Cost not found for job type {job_type}')
 
 
+def _get_cost_from_definition(job, cost_definition):
```
jtherrmann, line 130:
```suggestion
def _get_cost_from_definition(job: dict, cost_definition: dict) -> Decimal:
```

tests/test_dynamo/test_jobs.py
```diff
@@ -228,6 +245,73 @@ def test_get_credit_cost():
         dynamo.jobs._get_credit_cost({'job_type': 'INSAR_ISCE_BURST', 'job_parameters': {'foo': 'bar'}}, costs) == 1.0
     )
     assert dynamo.jobs._get_credit_cost({'job_type': 'INSAR_ISCE_BURST', 'job_parameters': {}}, costs) == 1.0
+    multi_burst_job = {
+        'job_type': 'INSAR_ISCE_MULTI_BURST',
+        'job_parameters': {'reference': ['g1', 'g2', 'g3'], 'looks': '5x1'},
+    }
+    assert dynamo.jobs._get_credit_cost(multi_burst_job, costs) == 10.0
```
jtherrmann, line 252: Let's add two more test cases, for `reference` length 1 and 2.

lib/dynamo/dynamo/jobs.py
williamh890: This doesn't work because it will always pull the first cost table it sees, regardless of it's parameter value.

job_spec/INSAR_ISCE_MULTI_BURST.yml
```diff
@@ -48,7 +48,110 @@ INSAR_ISCE_MULTI_BURST:
...
+      cost_parameter: looks
+      cost_table:
+        - parameter_value: 20x4
+          cost_table:
+            cost_parameter: reference
+            cost_table:
+              - parameter_value: 1
+                cost: 1.0
+              - parameter_value: 2
+                cost: 1.0
```
jtherrmann, line 60: Hmm, I'm realizing it's confusing that `cost_table` can either be a list *or* a dict. For example, the rendered `costs.json` looks like:

```json
[
  {
    "job_type": "INSAR_ISCE_MULTI_BURST",
    "cost_parameter": "looks",
    "cost_table": [
      {
        "parameter_value": "20x4",
        "cost_table": {
          "cost_parameter": "reference",
          "cost_table": [
            {
              "parameter_value": 1,
              "cost": 1.0
            },
            {
              "parameter_value": 2,
              "cost": 1.0
            },
...
```

I have a feeling this is also making the recursive function a bit confusing. But I'm not seeing the obvious solution at the moment.

We might also reconsider trying the flat lookup table approach with `(looks, length)` keys, e.g. `(20x4, 1), (20x4, 2), ... (20x4, 15)` if the recursive table seems too complicated.

job_spec/RTC_GAMMA.yml
```diff
@@ -88,14 +88,11 @@ RTC_GAMMA:
...
       cost_table:
-        - parameter_value: 30.0
-          cost: 5.0
-        - parameter_value: 20.0
-          cost: 15.0
-        - parameter_value: 10.0
-          cost: 60.0
+        30.0: 5.0
+        20.0: 15.0
+        10.0: 60.0
```
jtherrmann, line 95: This is where the float vs. int issue might bite us, I'd suggest using int keys but we should test submitting jobs with both float and int values for `resolution` and see what happens.
```suggestion
        30: 5.0
        20: 15.0
        10: 60.0
```

lib/dynamo/dynamo/jobs.py
```diff
@@ -117,22 +117,35 @@ def _prepare_job_for_database(
...
+    cost_table = cost_definition['cost_table']
+
+    for cost_parameter in cost_definition['cost_parameters']:
+        parameter_value = job['job_parameters'][cost_parameter]
+
+        if isinstance(parameter_value, list):
+            parameter_value = len(parameter_value)
+
+        if isinstance(parameter_value, float):
+            parameter_value = int(parameter_value)
```
jtherrmann, line 141: Oh OK, I see you're already handling the float -> int conversion. We probably want to verify that `parameter_value == int(parameter_value)` to make sure the float value doesn't have a fractional part. Though, we may also want to consider just changing how the resolution parameter works for `RTC_GAMMA`. Ideally I'd probably just convert it to an integer in the API spec, but I think that would break backwards compatibility. Something to think about more later.

jtherrmann (commented):
I haven't given the cost function a thorough read yet, but the cost definitions look like I'd expect.

williamh890:
Running validate only jobs gives these credit costs on multi-burst-sandbox

```
looks 5x1, ref: 1 credits: 1
looks 5x1, ref: 2 credits: 5
looks 5x1, ref: 3 credits: 10
looks 5x1, ref: 4 credits: 15
looks 5x1, ref: 5 credits: 20
looks 5x1, ref: 6 credits: 25
looks 5x1, ref: 7 credits: 30
looks 5x1, ref: 8 credits: 35
looks 5x1, ref: 9 credits: 40
looks 5x1, ref: 10 credits: 45
looks 5x1, ref: 11 credits: 90
looks 5x1, ref: 12 credits: 95
looks 5x1, ref: 13 credits: 100
looks 5x1, ref: 14 credits: 105
looks 5x1, ref: 15 credits: 110
looks 10x2, ref: 1 credits: 1
looks 10x2, ref: 2 credits: 1
looks 10x2, ref: 3 credits: 1
looks 10x2, ref: 4 credits: 5
looks 10x2, ref: 5 credits: 5
looks 10x2, ref: 6 credits: 5
looks 10x2, ref: 7 credits: 5
looks 10x2, ref: 8 credits: 5
looks 10x2, ref: 9 credits: 5
looks 10x2, ref: 10 credits: 10
looks 10x2, ref: 11 credits: 10
looks 10x2, ref: 12 credits: 10
looks 10x2, ref: 13 credits: 10
looks 10x2, ref: 14 credits: 10
looks 10x2, ref: 15 credits: 10
looks 20x4, ref: 1 credits: 1
looks 20x4, ref: 2 credits: 1
looks 20x4, ref: 3 credits: 1
looks 20x4, ref: 4 credits: 1
looks 20x4, ref: 5 credits: 5
looks 20x4, ref: 6 credits: 5
looks 20x4, ref: 7 credits: 5
looks 20x4, ref: 8 credits: 5
looks 20x4, ref: 9 credits: 5
looks 20x4, ref: 10 credits: 5
looks 20x4, ref: 11 credits: 5
looks 20x4, ref: 12 credits: 5
looks 20x4, ref: 13 credits: 5
looks 20x4, ref: 14 credits: 5
looks 20x4, ref: 15 credits: 5
```

williamh890:
Running `looks 5x1, ref: 15 credits: 110` without `validate_only` removed the correct number of credits from my accounts. Also `RTC_GAMMA` works correctly for different resolution parameters

job_spec/INSAR_ISCE_MULTI_BURST.yml
```diff
@@ -47,9 +47,58 @@ INSAR_ISCE_MULTI_BURST:
...
+          7: 30.0
+          8: 35.0
+          9: 40.0
+          10: 45.0
+          11: 90.0
+          12: 95.0
+          13: 100.0
+          14: 105.0
+          15: 110.0
+    EDC:
```
jtherrmann, line 101:
```suggestion
    EDC:
      cost_parameters: [looks, reference]
      cost_table:
        20x4:
          1: 1.0
          2: 1.0
          3: 1.0
          4: 1.0
          5: 5.0
          6: 5.0
          7: 5.0
          8: 5.0
          9: 5.0
          10: 5.0
          11: 5.0
          12: 5.0
          13: 5.0
          14: 5.0
          15: 5.0
        10x2:
          1: 1.0
          2: 1.0
          3: 1.0
          4: 5.0
          5: 5.0
          6: 5.0
          7: 5.0
          8: 5.0
          9: 5.0
          10: 10.0
          11: 10.0
          12: 10.0
          13: 10.0
          14: 10.0
          15: 10.0
        5x1:
          1: 1.0
          2: 5.0
          3: 10.0
          4: 15.0
          5: 20.0
          6: 25.0
          7: 30.0
          8: 35.0
          9: 40.0
          10: 45.0
          11: 90.0
          12: 95.0
          13: 100.0
          14: 105.0
          15: 110.0
    DEFAULT:
```

apps/render_cf.py
```diff
@@ -213,13 +213,11 @@ def render_default_params_by_job_type(job_types: dict) -> None:
...
-        {
-            'job_type': job_type,
-            **job_spec['cost_profiles'][cost_profile],
-        }
-        for job_type, job_spec in job_types.items()
-    ]
+    costs = {}
+
+    for job_type, job_spec in job_types.items():
+        costs[job_type] = job_spec['cost_profiles'][cost_profile]
```
jtherrmann, line 219: Can simplify even further:
```suggestion
    costs = {job_type: job_spec['cost_profiles'][cost_profile] for job_type, job_spec in job_types.items()}
```

lib/dynamo/dynamo/jobs.py
jtherrmann: Nevermind, I was overcomplicating this. What you have here is probably good.

lib/dynamo/dynamo/jobs.py
```diff
@@ -115,23 +115,39 @@ def _prepare_job_for_database(
...
+
+        try:
+            cost_lookup = cost_lookup[str(parameter_value)]
+        except KeyError:
+            raise ValueError(
+                f'Cost not found for job type {job["job_type"]} with {cost_parameter} == {parameter_value}'
+            )
+
+    return Decimal(cost_lookup)
```
jtherrmann, line 150: Should already be a `Decimal` because of how `COSTS` is defined at the beginning of the file, but would probably be a good idea to verify that here:
```suggestion
    if not isinstance(cost_lookup, Decimal):
        raise ValueError(f'Job type {job["job_type"]} has non-Decimal cost value')
    return cost_lookup
```

lib/dynamo/dynamo/jobs.py
```diff
@@ -115,23 +115,39 @@ def _prepare_job_for_database(
...
+        parameter_value = job['job_parameters'][cost_parameter]
+
+        if isinstance(parameter_value, list):
+            parameter_value = len(parameter_value)
+
+        if isinstance(parameter_value, float):
+            parameter_value = int(parameter_value)
+
+        try:
+            cost_lookup = cost_lookup[str(parameter_value)]
```
jtherrmann, line 144: I'd suggest factoring out a `_get_cost_parameter_value` helper function to help keep the logic for serializing parameter values self-contained:

```python
def _get_cost_parameter_value(job: dict, cost_parameter: str) -> str:
    parameter_value = job['job_parameters'][cost_parameter]

    if isinstance(parameter_value, list):
        parameter_value = len(parameter_value)

    if isinstance(parameter_value, float):
        parameter_value = int(parameter_value)
    
    return str(parameter_value)
```

lib/dynamo/dynamo/jobs.py
```diff
@@ -115,23 +115,57 @@ def _prepare_job_for_database(
...
+
+        try:
+            cost_lookup = cost_lookup[parameter_value]
+        except KeyError:
+            raise ValueError(
+                f'Cost not found for job type {job["job_type"]} with {cost_parameter} == {parameter_value}'
+            )
+
+    if not isinstance(cost_lookup, Decimal):
+        raise ValueError(f'Job type {job["job_type"]} has non-Decimal cost value')
```
jtherrmann, line 145: We should add a unit test for this error.

lib/dynamo/dynamo/jobs.py
```diff
@@ -115,23 +115,59 @@ def _prepare_job_for_database(
...
+        cost_parameter_value = str(int(parameter_value))
+
+    elif isinstance(parameter_value, list):
+        cost_parameter_value = str(len(parameter_value))
+
+    else:
+        raise ValueError(
+            f'Cost parameter {cost_parameter} for job type {job["job_type"]} has '
+            f'unsupported type {type(parameter_value)}'
+        )
```
jtherrmann, line 168: We should also add a unit test for this error.

apps/render_cf.py
```diff
@@ -242,11 +237,50 @@ def validate_job_spec(job_type: str, job_spec: dict) -> None:
...
+        if profile.keys() not in ({'cost_parameters', 'cost_table'}, {'cost'}):
+            raise ValueError(f'Cost definition for job type {job_type} has invalid keys: {profile.keys()}')
+
+        if 'cost_table' in profile:
+            if not isinstance(profile['cost_parameters'], list):
+                raise ValueError(
+                    f'Cost definition for job type {job_type} has invalid cost_parameters: Must be a list of strings.'
+                )
+
+            validate_cost_table(profile['cost_table'], job_type)
```
jtherrmann, line 250: We *could* update `test_validate_job_spec` in `test_render_cf.py`, but I'm not super concerned.

apps/render_cf.py
```diff
@@ -242,11 +237,50 @@ def validate_job_spec(job_type: str, job_spec: dict) -> None:
...
             raise ValueError(f'{job_type} has image {step["image"]} but docker requires the image to be all lowercase')
 
 
+def validate_cost_table(cost_table: dict | float, job_type: str) -> None:
+    if isinstance(cost_table, dict):
+        for key, value in cost_table.items():
+            if not isinstance(key, str) and not isinstance(key, int):
+                raise ValueError(
+                    f'Cost definition for job type {job_type} has invalid cost_table: '
+                    f'all cost_table keys must be strings or ints {type(key)}.'
```
jtherrmann, line 263:
```suggestion
                    f'all cost_table keys must be strings or ints, but {key} has type {type(key)}.'
```

apps/render_cf.py
williamh890: Good idea, I'll add em

apps/render_cf.py
```diff
@@ -242,11 +237,50 @@ def validate_job_spec(job_type: str, job_spec: dict) -> None:
...
+                )
+
+            return validate_cost_table(value, job_type)
```
jtherrmann, line 266:
```suggestion
            validate_cost_table(value, job_type)
```

apps/render_cf.py
```diff
@@ -242,11 +237,50 @@ def validate_job_spec(job_type: str, job_spec: dict) -> None:
...
+                'all cost_table costs must be whole numbers.'
+            )
+
+        return
+
+    else:
+        raise ValueError(
+            f'Cost definition for job type {job_type} has invalid cost_table: '
+            f'Cost table must be a nested dictionary of costs. {type(cost_table)}'
+        )
```
jtherrmann, line 281: Cool, I didn't know about `float.is_integer`. But we actually *don't* want to require that credit costs are integer values, we intentionally made them `Decimal` rather than `int` in case we ever want to support credit costs less than 1.
```suggestion
            validate_cost_table(value, job_type)

    elif not isinstance(cost_table, float):
        raise ValueError(
            f'Cost definition for job type {job_type} has invalid cost_table: '
            f'Cost table must be a nested dictionary of costs, got type {type(cost_table)}'
        )
```

apps/render_cf.py
```diff
@@ -242,11 +237,49 @@ def validate_job_spec(job_type: str, job_spec: dict) -> None:
                 f'but should have {expected_param_fields}'
             )
 
+    for profile in job_spec['cost_profiles'].values():
+        if profile.keys() not in ({'cost_parameters', 'cost_table'}, {'cost'}):
+            raise ValueError(f'Cost definition for job type {job_type} has invalid keys: {profile.keys()}')
+
+        if 'cost_table' in profile:
+            if not isinstance(profile['cost_parameters'], list):
+                raise ValueError(
```
jtherrmann, line 246: We might be missing a test case for the `invalid cost_parameters` error.

job_spec/INSAR_ISCE_MULTI_BURST.yml
jtherrmann, line 1: I confirmed that the cost table here matches [average-credit-costs.csv](https://github.com/ASFHyP3/scratch/blob/main/multi-burst-scaling/credits/spot-prices/average-credit-costs.csv) for 20x4 and 10x2, and [adjusted-5x1-credits.csv](https://github.com/ASFHyP3/scratch/blob/main/multi-burst-scaling/credits/spot-prices/adjusted-5x1-credits.csv) for 5x1.

jtherrmann (approved):

PR #2642, opened 2025-03-06
Finish adding type annotations

jtherrmann:
TODO:

- [x] Merge https://github.com/ASFHyP3/hyp3/pull/2631, then change base branch for this PR to `develop`


lib/lambda_logging/lambda_logging/__init__.py
```diff
@@ -11,11 +13,11 @@ class UnhandledException(Exception):
     pass
 
 
-def log_exceptions(lambda_handler):
+def log_exceptions[T](lambda_handler: Callable[[dict, Any], T]) -> Callable[[dict, Any], T]:
```
forrestfwilliams, line 16: `[T]` seems like a strange thing to include in a function name. Is this a typing thing?

jtherrmann: Yep! Generics: https://docs.python.org/3/whatsnew/3.12.html#pep-695-type-parameter-syntax

In this case it just means `T` is whatever type is returned by the `lambda_handler` function that's being wrapped by the `log_exceptions` decorator.

forrestfwilliams (approved):

PR #2676, opened 2025-03-19
Report HyP3 API version

jhkennedy:
Moving deployments and their config out of this repo will likely introduce a lag between release and deployment, but there's no way to check that version of any particular HyP3 deployment from the API itself since we don't report it.

**Critically**, @ASFHyP3/tools , in order to start reporting version numbers, we'll need to change up how the DAAC deployments are done. Currently, a version number is only tagged *after* a successful production EDC deployment. That will need to be reversed, and with multiple teams developing and releasing features, there's no one canonical production deployment anymore, so the original motivation for this is moot, IMO.

This:
* adds a method to `render-cf.py` to get an API version number from [setuptools_scm](https://setuptools-scm.readthedocs.io/en/latest/)

For example, if you were on 40bb4ed8, which is currently the HEAD of this PR, the version numbers would look like: `10.1.0.dev3+g40bb4ed8`, however, if you're on a tagged commit, you will just get the version number like `10.2.1`.
  
  **Importantly**, the version number will not be reported correctly with shallow clones or clones where the tags haven't been pulled. However, shallow clones are still possible if you set the `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_HYP3` environment variable.
  
* Updates our OpenAPI Spec and Swagger UI to report the version number
* breaks the version tagging step out into its own deployment, triggered off of pushes to main
* Breaks the DAAC deployment maturities out into their own workflows
* triggers all production-level deployments off of pushes of `v*` tags


.github/workflows/deploy-daac-test.yml
```diff
@@ -1,10 +1,9 @@
-name: Deploy DAAC Stacks to AWS
+name: Deploy DAAC test to AWS
 
 on:
   push:
-    branches:
-      - main
-      - develop
+    tags:
+      - 'v*'
```
asjohnston-asf, line 6: `develop` for the test environment, yes?
```suggestion
    branches:
      - develop
```

jhkennedy: Oh woops!

asjohnston-asf:
I feel triggering production deployments on [release](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release) might be a better approach than on [tag](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs)? And then pass `GITHUB_REF` to the render script to determine the API version?

jhkennedy:
> I feel triggering production deployments on [release](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release) might be a better approach than on [tag](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs)? And then pass GITHUB_REF to the render script to determine the API version?

Maybe? We do use both:
* On `v*` tag: https://github.com/search?q=org%3AASFHyP3%20path%3A.github%2Fworkflows%2F*.yml%20%22v*%22&type=code
* on release: https://github.com/search?q=org%3AASFHyP3+path%3A.github%2Fworkflows%2F*.yml+%22release%3A%22&type=code

I generally think of the tag as the canonical representation of the version, and the release is how we communicate it, so slightly prefer triggering off the tag, but either are viable.

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [10.2.0 ]
```
jtherrmann, line 7:
```suggestion
## [10.2.0]
```

jhkennedy: :facepalm: This is the 2nd time I've done this... I wonder if PyCharm or something is doing this to me.

jtherrmann (dismissed):

jtherrmann (approved):

jtherrmann:
Validated with a local API:
- Confirmed I see `10.2.2.dev28+g17321fa4` when on `17321fa4`, in the Swagger UI, page title, and `/openapi.json`
- Confirmed I see `10.2.2` if I tag the commit with `v10.2.2`

PR #2690, opened 2025-03-26
Replace uses of `Any` where appropriate

jtherrmann:


apps/api/src/hyp3_api/routes.py
```diff
@@ -104,10 +103,10 @@ def default(self, o: object) -> object:
 
 
 class CustomJSONProvider(JSONProvider):
-    def dumps(self, obj: Any, **kwargs: Any) -> str:
+    def dumps(self, obj: object, **kwargs: object) -> str:
```
jhkennedy, line 106: :thinking: why use object here? kwargs is generally a `dict` and `object` is the base class of all python objects so is (superficially) similar to `Any`...

jtherrmann: Oh yeah, according to https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values the annotation is the type of the arguments themselves, not the `kwargs` dict.

And with regard to object vs. Any, object is preferred because it's a static type and so will be type checked, whereas Any is just the dynamic type that bypasses all static type checking: https://mypy.readthedocs.io/en/stable/dynamic_typing.html#any-vs-object

jhkennedy (approved):

PR #2692, opened 2025-03-27
Rename jobs

jtherrmann:
Changes have been deployed and tested at https://hyp3-multi-burst-sandbox.asf.alaska.edu

TODO:

- [x] Decide how to handle empty request body
- [x] TODO/FIXME in code
- [x] Tests
- [x] Final sandbox validation before merging (API behavior and Swagger UI rendering)
- [x] Changelog


lib/dynamo/dynamo/jobs.py
```diff
@@ -232,6 +235,36 @@ def update_job(job: dict) -> None:
...
+            UpdateExpression=update_expression,
+            ConditionExpression='user_id = :user_id',
+            ExpressionAttributeValues={':user_id': user_id, **name_value},
+            ExpressionAttributeNames={'#name': 'name'},
+            ReturnValues='ALL_NEW',
+        )['Attributes']
+    except botocore.exceptions.ClientError as e:
+        if e.response['Error']['Code'] == 'ConditionalCheckFailedException':
+            raise UpdateJobForDifferentUserError("You cannot modify a different user's job")
+        raise
```
williamh890, line 263:
```suggestion
        raise e
```

jtherrmann: Bare `raise` already re-raises the current exception, so the same as `raise e` in this case. I'd prefer to keep it as-is for consistency with the functions in [dynamo.user](https://github.com/ASFHyP3/hyp3/blob/develop/lib/dynamo/dynamo/user.py), which also follow this pattern.

Though, I should probably add a unit test for the `raise` case. I recall running into difficulty when trying to add a test for this last week, but I don't remember what the issue was, so I'll try again.

jtherrmann: Added a unit test for the `raise` statement.

williamh890: Cool, didn't know that. I though you needed to pass the exception.

williamh890 (approved):

PR #2724, opened 2025-04-24
Add OPERA_RTC job type

jtherrmann:
In addition to the changes described in the Changelog, this PR also includes some improvements to our tests:
- Makes it easier to run `pytest` outside of the `Makefile`.
- Allows marking tests that require a network connection so they can be skipped if desired.
- Fixes a bug in `tests/test_api/test_submit_job.py` in which `responses` mocking was not being cleaned up after use, which blocked the new OPERA RTC integration tests from making actual CMR requests.
- Moves utils that are specific to `tests/test_api/test_submit_job.py` out of `tests/test_api/conftest.py`; this was motivated by the bug described above and a desire to make it more obvious which utils are being used by which test files.
- Renames some test util functions to make it more explicit that they are mocking CMR responses.
- Explicitly import `pytest` for `pytest.raises` in `tests/test_api/test_validation.py`
- Remove the burst InSAR test cases from `test_validate_jobs` because they didn't add much value and because they were not mocking their CMR responses, and the original intent for that test function was to mock all responses. See the commit message for https://github.com/ASFHyP3/hyp3/pull/2724/commits/aab217bc1c6f299d32ffad401a5fa18803fedd78

TODO:

- [x] https://github.com/ASFHyP3/hyp3/pull/2731
- [x] Disallow older granules (instead of checking IPF version); see https://github.com/ASFHyP3/hyp3/issues/2739 for context
- [x] TODO/FIXME in code
- [x] Unit tests
- [x] Integration tests for API validation for this job type; see https://github.com/ASFHyP3/hyp3/issues/2740
- [x] https://github.com/ASFHyP3/hyp3/issues/2742
- [x] Sandbox testing
- [x] Determine credit cost after initial approval (but before merging?)
- [x] Changelog


apps/api/src/hyp3_api/validation.py
```diff
@@ -209,6 +216,51 @@ def check_bounding_box_size(job: dict, _, max_bounds_area: float = 4.5) -> None:
...
+    if not _has_opera_rtc_static_coverage(granule):
+        raise GranuleValidationError(f'Granule {granule} is outside of the OPERA RTC processing extent.')
+
+
+def check_opera_rtc_date(job: dict, _) -> None:
+    granules = job['job_parameters']['granules']
+    if len(granules) != 1:
+        raise InternalValidationError(f'Expected 1 granule, got {granules}')
+
+    granule = granules[0]
```
williamh890, line 246: `granules` as a hedge for the future feels strange to me.

jtherrmann: Yep, fair. @forrestfwilliams said there might be a 5% chance we would allow multiple granules with this job type in the future, so I think I'm reluctant to intentionally restrict ourselves to a single granule by changing the job spec parameter to a `granule` string. The experience with attempting to convert our single-burst InSAR job type to multi-burst InSAR and realizing that we needed to break backwards compatibility has made me slightly more in favor of future-proofing ourselves in these situations. But I'm on the fence and could be convinced.

williamh890 (approved):

README.md
```diff
@@ -28,10 +28,26 @@ A processing environment for HyP3 Plugins in AWS.
...
    ```
-3. Run the tests
+3. Run the tests:
    ```
    make tests
    ```
+   Alternatively, you can invoke `pytest` directly (e.g. for passing command-line arguments):
+   ```
+   eval $(make env)
+   make render && pytest
```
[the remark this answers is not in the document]
jtherrmann: I suspect so, but I still like having the `eval $(make env)` option because I can run it once and then run other utilities such as `ipython`.

README.md
```diff
@@ -28,10 +28,26 @@ A processing environment for HyP3 Plugins in AWS.
...
+   ```
+   In particular, to skip tests that require a network connection, run:
+   ```
+   pytest -m 'not network'
+   ```
+   And to run *only* those tests:
+   ```
+   pytest -m network
```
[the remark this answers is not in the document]
jtherrmann: They always run during GH Actions, this just gives the option for skipping them when developing locally, in case you internet connection is slow or there's a CMR outage.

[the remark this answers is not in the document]
jtherrmann: Apparently a common workflow is to run fast/unit tests first, and then only run slow/integration tests if the fast tests succeed.

job_spec/OPERA_RTC_S1.yml
```diff
@@ -0,0 +1,45 @@
...
+      cost: 1.0
+    DEFAULT:
+      cost: 1.0
+  validators:
+    - check_opera_rtc_s1_date
+    - check_dem_coverage
+    # Static coverage check is the slowest because it involves a CMR query, so should come last
+    - check_opera_rtc_s1_static_coverage
+  steps:
+    - name: ''
```
williamh890, line 31: What's the reason for the name with empty string? I see other job types have this, while others have names.

williamh890 (approved):

job_spec/OPERA_RTC_S1.yml
jtherrmann: From what I remember, we decided on this convention when implementing the ability for a job spec to have multiple steps. The name of the job type gets prefixed onto the name of the step, e.g. `OPERA_RTC_S1_STEP1`. If you just want the name of the step to be the same as the name of the job, then you leave name empty.

PR #2731, opened 2025-04-28
first draft of opera rtc extent validator

asjohnston-asf:


apps/api/src/hyp3_api/validation.py
```diff
@@ -215,6 +215,24 @@ def check_bounding_box_size(job: dict, _, max_bounds_area: float = 4.5) -> None:
...
+        'granule_ur': f'OPERA_L2_RTC-S1-STATIC_T*-{burst_id}-{swath}_*',
+        'options[granule_ur][pattern]': 'true',
+    }
+    response = requests.get(CMR_URL, params=params)
+    response.raise_for_status()
+    return len(response.json()['feed']['entry']) > 0
+
+
+def check_rtc_static_coverage(_, granule_metadata: list[dict]) -> None:
+    bad_granules = [g['name'] for g in granule_metadata if not _has_rtc_static_coverage(g['name'])]
```
jtherrmann, line 231: I don't see a `name` field in the CMR record for these granules (e.g. [this one](https://cmr.earthdata.nasa.gov/search/granules.umm_json?short_name=SENTINEL-1_BURSTS&granule_ur=S1_073251_IW2_20250413T020809_VV_EF1E-BURST)) but we can get this from the job dict itself. Also, I might take back what I said this morning about assuming one granule; it seems easy enough to support multiple granules and that would allow for the slight chance that we ever want to support multiple granules for this job type.
```suggestion
def check_rtc_static_coverage(job: dict, _) -> None:
    bad_granules = [granule for granule in job['job_parameters']['granules'] if not _has_rtc_static_coverage(granule)]
```

apps/api/src/hyp3_api/validation.py
```diff
@@ -215,6 +215,24 @@ def check_bounding_box_size(job: dict, _, max_bounds_area: float = 4.5) -> None:
         )
 
 
+def _has_rtc_static_coverage(granule_name: str) -> bool:
+    burst_id, swath = granule_name.split('_')[1:3]
```
jtherrmann, line 219: See my pending question to @forrestfwilliams in Mattermost regarding the term "burst ID".

tests/test_api/test_validation.py
```diff
@@ -479,6 +479,26 @@ def test_check_bounding_box_size():
         validation.check_bounding_box_size(job, None, max_bounds_area=99.9)
 
 
+def test_has_rtc_static_coverage():
```
jtherrmann, line 482: If we're not mocking these then I'd be concerned about the runtime of the test suite when making all of these CMR queries. I'd prefer two test cases (one granule with coverage, one without) so that the test only has to run two queries.

And perhaps would be good to test the `check_rtc_static_coverage` function itself, rather than the helper function.

jtherrmann: Ran the test locally. It's faster than I thought, though still noticeably slower than the other tests, so still might be nice to reduce the number of assertions.

apps/api/src/hyp3_api/validation.py
jtherrmann: Oh, didn't realize we add the `name` field as part of the `_get_cmr_metadata` function.

apps/api/src/hyp3_api/validation.py
jtherrmann: Forrest says burst ID = burst number + swath, so we should change the variable name here to `burst_number`.

apps/api/src/hyp3_api/validation.py
jtherrmann: But I still might recommend using the granules from the job parameters, to reduce our reliance on CMR metadata given recent discussions about switching to `umm_json` at some point.

jtherrmann (approved):

Issue #2740, opened 2025-04-29
Add integration tests for HyP3 API job validation


jtherrmann:
TODO: description

jtherrmann:
See [`tests/test_api/test_opera_rtc_s1.py`](https://github.com/ASFHyP3/hyp3/blob/develop/tests/test_api/test_opera_rtc_s1.py) for an example of how to write these tests. Make sure to mark them with `@pytest.mark.network` (see the [README](https://github.com/ASFHyP3/hyp3?tab=readme-ov-file#developer-setup)).

Issue #2737, opened 2025-04-29
Validate granules using UMM_JSON metadata from CMR


jtherrmann:
TODO: description

PR #2772, opened 2025-05-23
ARIA_S1_GUNW Jobs use reference and secondary dates as inputs

williamh890:
Refactors:
- Switch to generic `ValidationError` Exception type in `validation.py`
- Skip downloading job granule metadata if no jobs contain granules

TODO: 
- [x] Add tests
- [x] Add integration tests (error messages + openapi schema checks)
- [x] Refactor validation exceptions and add tests for them
- [x] Add good example values
- [ ] Check that example dates pulled from develop works (maybe after merge?)
- [x] Wait for container pr to be merged https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/pull/207
- [ ] Wait for `test` image to build: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/actions/runs/15619654840


jtherrmann (commented):
I think ideally we would add integration tests for the new validator, per https://github.com/ASFHyP3/hyp3/issues/2740 (see [`test_opera_rtc_s1.py`](https://github.com/ASFHyP3/hyp3/blob/develop/tests/test_api/test_opera_rtc_s1.py) for some examples), but I'll leave it to your judgment whether that's a priority for this PR. The main benefit is that it will allow us to refactor our validation architecture as needed while remaining confident that we haven't broken our existing job types (e.g. for https://github.com/ASFHyP3/hyp3/issues/2737), but we could also just wait to add integration tests if/when we do that refactoring.

Edit: Maybe I'd advocate a bit more strongly for them after finding a bug that would have been caught be integration testing (see below).

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [10.6.1]
```
jtherrmann, line 7: It would be `10.6.0` for a minor release, but technically I think this is a breaking change for a job type that's already available in edc-prod, so technically it should be a major release. But I'm open to push-back since I don't think we've really advertised this job type yet?

apps/api/src/hyp3_api/validation.py
```diff
@@ -26,6 +26,10 @@ class GranuleValidationError(Exception):
     pass
 
 
+class DateValidationError(Exception):
```
jtherrmann, line 29: If you want to add a new `DateValidationError` exception type, then it will need to be handled in [`handlers.post_jobs`](https://github.com/ASFHyP3/hyp3/blob/develop/apps/api/src/hyp3_api/handlers.py#L34), otherwise it will just result in an Internal Server Error. (This is something that integration tests would catch!)

On the other hand, I don't think that having different types of validation exceptions actually provides any benefit, since they all show up as `400 Bad Request` for the user, but I haven't gotten around to refactoring to combine the three existing types (`BoundsValidationError, GranuleValidationError, MultiBurstValidationError`) into a single exception type. So I'm fine with adding a new exception type for now.

tests/test_api/test_validation.py
```diff
@@ -582,3 +583,62 @@ def test_check_opera_rtc_s1_date_max_configurable(monkeypatch):
         validation.check_opera_rtc_s1_date(
             {'job_parameters': {'granules': ['S1_000000_IW1_20250506T000000_VV_0000-BURST']}}, None
         )
+
+
+def test_check_aria_s1_gunw_dates(monkeypatch):
```
jtherrmann, line 588: I'm not seeing `monkeypatch` used in this test:
```suggestion
def test_check_aria_s1_gunw_dates():
```

apps/api/src/hyp3_api/validation.py
```diff
@@ -241,6 +245,37 @@ def check_opera_rtc_s1_static_coverage(job: dict, _) -> None:
...
+                f'"{job_date_key}" is {job_date} which is before the start of the sentinel 1 mission ({s1_start_date}).'
+            )
+
+    if job_dates['reference_date'] == job_dates['secondary_date']:
+        raise DateValidationError(
+            f'reference and secondary dates are equal, must be different dates ({job_dates["reference_date"]}).'
+        )
+
+    if job_dates['reference_date'] > job_dates['secondary_date']:
+        raise DateValidationError('reference date must be earlier than secondary date.')
```
asjohnston-asf, line 280: We want reference to be the later date, and there's no need for a separate equality check:
```suggestion
    if not job_dates['secondary_date'] < job_dates['reference_date']:
        raise DateValidationError('secondary date must be earlier than reference date.')
```

williamh890:
> I think ideally we would add integration tests for the new validator, per #2740 (see [`test_opera_rtc_s1.py`](https://github.com/ASFHyP3/hyp3/blob/develop/tests/test_api/test_opera_rtc_s1.py) for some examples), but I'll leave it to your judgment whether that's a priority for this PR. The main benefit is that it will allow us to refactor our validation architecture as needed while remaining confident that we haven't broken our existing job types (e.g. for #2737), but we could also just wait to add integration tests if/when we do that refactoring.
> 
> Edit: Maybe I'd advocate a bit more strongly for them after finding a bug that would have been caught by integration testing (see below).

Is the idea to do all the tests for the validator at the integration layer instead? So moving checks `test_check_aria_s1_gunw_dates` into it's own file using the integration test format?

tests/test_api/test_aria_s1_gunw.py
```diff
@@ -0,0 +1,46 @@
...
+                    'job_parameters': {'reference_date': '2022-01-01', 'secondary_date': '2022-01-02', 'frame_id': 1}
+                }
+            ],
+        },
+    )
+    assert response.status_code == HTTPStatus.OK
+    assert len(tables.jobs_table.scan()['Items']) == 1
+
+
+def test_aria_job_with_bad_date(client, tables, approved_user):
```
jtherrmann, line 30: I believe this is still a networked test because the validation code queries CMR before running any of the validators. Only the OpenAPI spec (e.g. string regexes) specified in the job spec is applied before the CMR query.
```suggestion
@pytest.mark.network
def test_aria_job_with_bad_date(client, tables, approved_user):
```

job_spec/ARIA_S1_GUNW.yml
```diff
@@ -1,38 +1,25 @@
...
-          pattern: "^S1[AB]_IW_SLC__1S[SD]V"
-          minLength: 67
-          maxLength: 67
-          example: S1A_IW_SLC__1SDV_20250103T010137_20250103T010204_057273_070BB6_CD45
+        description: The date to find Sentinel-1 SLC granules to use as secondary scenes for InSAR processing
+        type: string
+        maxLength: 10
+        minLength: 10
+        pattern: "^[0-9]{4}-[0-9]{2}-[0-9]{2}$"
+        example: "2025-01-21"
```
jtherrmann, line 22: I notice the example value `2025-01-21` is the same as for the reference date, so it would fail validation? If so, we should add a real example that would serve as a nice default test job (i.e. runs quickly and produces a meaningful product).

job_spec/ARIA_S1_GUNW.yml
```diff
@@ -1,38 +1,25 @@
...
-          pattern: "^S1[AB]_IW_SLC__1S[SD]V"
-          minLength: 67
-          maxLength: 67
-          example: S1A_IW_SLC__1SDV_20250127T010136_20250127T010203_057623_07199D_4B63
-    secondary:
+        description: The date to find Sentinel-1 SLC granules to use as reference scenes for InSAR processing
+        type: string
+        maxLength: 10
+        minLength: 10
+        pattern: "^[0-9]{4}-[0-9]{2}-[0-9]{2}$"
```
jtherrmann, line 13: Slightly more concise:
```suggestion
        pattern: "^\d{4}-\d{2}-\d{2}$"
```

job_spec/ARIA_S1_GUNW.yml
```diff
@@ -1,38 +1,25 @@
...
-          type: string
-          pattern: "^S1[AB]_IW_SLC__1S[SD]V"
-          minLength: 67
-          maxLength: 67
-          example: S1A_IW_SLC__1SDV_20250103T010137_20250103T010204_057273_070BB6_CD45
+        description: The date to find Sentinel-1 SLC granules to use as secondary scenes for InSAR processing
+        type: string
+        maxLength: 10
+        minLength: 10
+        pattern: "^[0-9]{4}-[0-9]{2}-[0-9]{2}$"
```
jtherrmann, line 21:
```suggestion
        pattern: "^\d{4}-\d{2}-\d{2}$"
```

job_spec/ARIA_S1_GUNW.yml
jtherrmann: Nevermind, Andrew's on it.

job_spec/ARIA_S1_GUNW.yml
jtherrmann, line 1: We may also want to include the OpenAPI spec changes in the integration tests (e.g. date string format), to verify that invalid parameters trigger the `is not valid under any of the given schemas` error message. See [`test_opera_rtc_s1.test_opera_rtc_s1_vh`](https://github.com/ASFHyP3/hyp3/blob/develop/tests/test_api/test_opera_rtc_s1.py#L30) as an example.

job_spec/ARIA_S1_GUNW.yml
```diff
@@ -1,38 +1,25 @@
 ARIA_S1_GUNW:
   required_parameters:
-    - reference
-    - secondary
+    - reference_date
+    - secondary_date
     - frame_id
   parameters:
-    reference:
+    reference_date:
```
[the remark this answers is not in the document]
williamh890: I like the better error message so I think I'm going to leave the check in the validator.

job_spec/ARIA_S1_GUNW.yml
williamh890: Going with `format: date` so there will be no need for the pattern.

apps/api/src/hyp3_api/validation.py
```diff
@@ -236,9 +232,33 @@ def check_opera_rtc_s1_static_coverage(job: dict, _) -> None:
...
+    todays_date = date.today()
+
+    for job_date_key, job_date in job_dates.items():
+        if job_date > todays_date:
+            raise ValidationError(f'"{job_date_key}" is {job_date} which is a date in the future.')
+
+        if job_date < s1_start_date:
+            raise ValidationError(
+                f'"{job_date_key}" is {job_date} which is before the start of the sentinel 1 mission ({s1_start_date}).'
+            )
```
jtherrmann, line 258: It might be more clear to factor out a helper function for handling the code that has to be run for both dates, e.g. `_check_aria_s1_gunw_date(job_params: dict, date_key: str)`, and then call that function for both date keys.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - `ARIA_S1_GUNW` jobs now pass `--min-frame-coverage 0.01` to the `DockerizedTopsApp` container.
+- `ARIA_S1_GUNW` now takes `reference_date` and `secondary_date` as inputs instead of `reference` and `secondary` granule lists
```
jtherrmann, line 14: Will need its own changelog section.

apps/api/src/hyp3_api/validation.py
```diff
@@ -236,11 +232,44 @@ def check_opera_rtc_s1_static_coverage(job: dict, _) -> None:
...
+
+
+def check_secondary_before_reference_date(job: dict, _) -> None:
+    def format_date(key: str) -> date:
+        return date.fromisoformat(job['job_parameters'][key])
+
+    reference, secondary = format_date('reference_date'), format_date('secondary_date')
+
+    if secondary >= reference:
+        raise ValidationError('secondary date must be earlier than reference date.')
```
jtherrmann, line 252: Can we just combine these two validator functions along these lines?

```suggestion
def check_aria_s1_gunw_dates(job: dict, _) -> None:
    def format_date(key: str) -> date:
        return date.fromisoformat(job['job_parameters'][key])

    reference, secondary = format_date('reference_date'), format_date('secondary_date')
    _validate_date_during_s1('reference_date', reference)
    _validate_date_during_s1('secondary_date', secondary)

    if secondary >= reference:
        raise ValidationError('secondary date must be earlier than reference date.')
```

apps/api/src/hyp3_api/validation.py
```diff
@@ -236,11 +232,44 @@ def check_opera_rtc_s1_static_coverage(job: dict, _) -> None:
...
+        raise ValidationError(
+            f'"{date_name}" is {date_value} which is before the start of the sentinel 1 mission ({s1_start_date}).'
         )
 
 
+def _get_date_from_job(job_parameters: dict, date_param_key: str) -> date:
+    param_date_str = job_parameters[date_param_key]
+    return date.fromisoformat(param_date_str)
+
+
```
jtherrmann, line 272: Looks like this function is no longer used:
```suggestion
```

apps/api/src/hyp3_api/validation.py
williamh890: I kinda like it for testing, and it makes these validators a bit more generic but i don't mind combining them.

job_spec/ARIA_S1_GUNW.yml
```diff
@@ -47,7 +29,7 @@ ARIA_S1_GUNW:
     DEFAULT:
       cost: 1.0
   validators:
-    - check_dem_coverage
+    - check_aria_s1_gunw_dates
```
jtherrmann, line 32: I assume want to keep the DEM coverage check?
```suggestion
    - check_aria_s1_gunw_dates
    - check_dem_coverage
```

jtherrmann: Oh wait nevermind, I get it.

jtherrmann (approved):

PR #2788, opened 2025-06-03
Add min_frame_coverage to INSAR_ISCE job spec

jacquelynsmale:


jacquelynsmale:
This can be marked ready after https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/pull/208 is merged

job_spec/INSAR_ISCE.yml
```diff
@@ -84,6 +84,12 @@ INSAR_ISCE:
           - GMAO
           - HRES
           - HRRR
+    min_frame_coverage:
+      api_schema:
+        description: Minimum amount of common area the reference and secondary granules overlap. Default is 0.01.
+        default: 0.01
```
jtherrmann, line 90:
```suggestion
        default: 0.01  # Must be kept in sync with DockerizedTopsApp
```

job_spec/INSAR_ISCE.yml
```diff
@@ -84,6 +84,12 @@ INSAR_ISCE:
           - GMAO
           - HRES
           - HRRR
+    min_frame_coverage:
+      api_schema:
+        description: Minimum amount of common area the reference and secondary granules overlap. Default is 0.01.
```
jtherrmann, line 89: Something like this? Also, not sure we need to include the default value in the description.
```suggestion
        description: Minimum amount of the frame that must be covered by the overlap of the reference and secondary granules.
```

job_spec/INSAR_ISCE.yml
```diff
@@ -84,6 +84,13 @@ INSAR_ISCE:
           - GMAO
           - HRES
           - HRRR
+    min_frame_coverage:
+      api_schema:
+        description: Minimum amount of the frame that must be covered by the overlap of the reference and secondary granules.
```
jtherrmann, line 89:
```suggestion
        description: Minimum amount of the frame that must be covered by the overlap between the reference and secondary granules; standard ARIA products require this to be set to 0.9.
```

job_spec/INSAR_ISCE.yml
```diff
@@ -84,6 +84,13 @@ INSAR_ISCE:
           - GMAO
           - HRES
           - HRRR
+    min_frame_coverage:
```
jtherrmann, line 87: And let's move the new parameter to right under the `frame_id` param, to match the container CLI.

jtherrmann (approved):

PR #2790, opened 2025-06-04
Add mypy to our Makefile

jhkennedy:
I was surprised when I ran `make static` that mypy was not run, as all our other static analyses are run at that time.

This:
* Adds a mypy make target
* adds mypy to the static target

Marking this as bumpless (?) since we haven't stated make changes since hyp3 v2.8.0.


jtherrmann (approved):
Good catch, I forgot `make static` was a thing.

PR #2802, opened 2025-06-11
Support EDL token authentication

asjohnston-asf:
# TODO
- [ ] support both cookie-based and token-based authentication
- [ ] avoid re-fetching the EDL signing key with every request
- [ ] figure out how `request.authorization` works so we don't have to parse the `Authorization` header ourselves
- [ ] add/update tests

### DIFF
--- a/apps/api/src/hyp3_api/auth.py
+++ b/apps/api/src/hyp3_api/auth.py
@@ -8,7 +8,9 @@ def decode_token(token: str | None) -> dict | None:
     if token is None:
         return None
     try:
-        return jwt.decode(token, environ['AUTH_PUBLIC_KEY'], algorithms=environ['AUTH_ALGORITHM'])
+        jwks_client = jwt.PyJWKClient('https://urs.earthdata.nasa.gov/.well-known/edl_ops_jwks.json')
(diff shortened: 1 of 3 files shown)


apps/api/src/hyp3_api/auth.py
```diff
@@ -4,13 +4,29 @@
...
+    except jwt.exceptions.InvalidTokenError as e:
+        raise InvalidTokenException(f'Invalid authorization token provided: {str(e)}')
+    return payload['uid'], token
+
+
+def decode_asf_cookie(cookie: str) -> tuple[str, str]:
+    try:
+        payload = jwt.decode(cookie, environ['AUTH_PUBLIC_KEY'], algorithms=environ['AUTH_ALGORITHM'])
+    except jwt.exceptions.InvalidTokenError as e:
+        raise InvalidTokenException(f'Invalid authorization cookie provided: {str(e)}')
```
jtherrmann, line 24: Are there any security concerns with exposing the error message directly from the `jwt` library by putting `str(e)` in the error message? I don't think there would be, since we're just doing public key decoding, but just want to make sure.

PR #2807, opened 2025-06-12
Add tests for auth error cases

williamh890:
TODO: 
- [x] Use magicmock and check called with correct params ([example](https://github.com/ASFHyP3/hyp3/blob/develop/tests/test_start_execution.py#L184))
- [x] refactor `get_mock_jwt_cookie` into conftest
- [x] Test for `route.py` (maybe)?

### DIFF
--- a/tests/test_api/test_api_spec.py
+++ b/tests/test_api/test_api_spec.py
@@ -1,6 +1,6 @@
 from http import HTTPStatus
 
-from test_api.conftest import AUTH_COOKIE, JOBS_URI, USER_URI, login, get_mock_jwt_cookie
+from test_api.conftest import AUTH_COOKIE, JOBS_URI, USER_URI, get_mock_jwt_cookie, login
 
 
 ENDPOINTS = {
--- a/tests/test_api/test_api_spec.py
+++ b/tests/test_api/test_api_spec.py
@@ -39,8 +39,8 @@ def test_invalid_cookie(client):
     for uri in ENDPOINTS:
         client.set_cookie(domain='localhost', key=AUTH_COOKIE, value='garbage I say!!! GARGBAGE!!!')
         response = client.get(uri)
-        assert 'Invalid authorization cookie provided' in response.json['detail']
         assert response.status_code == HTTPStatus.UNAUTHORIZED
+        assert 'Invalid authorization cookie provided' in response.json['detail']
 
 
 def test_invalid_bearer(client):
--- a/tests/test_api/test_api_spec.py
+++ b/tests/test_api/test_api_spec.py
@@ -1,7 +1,6 @@
 from http import HTTPStatus
 
-from hyp3_api import auth
-from test_api.conftest import AUTH_COOKIE, JOBS_URI, USER_URI, login
+from test_api.conftest import AUTH_COOKIE, JOBS_URI, USER_URI, login, get_mock_jwt_cookie
 
 
 ENDPOINTS = {
@@ -48,7 +47,7 @@ def test_expired_cookie(client):
         client.set_cookie(
             domain='localhost',
             key=AUTH_COOKIE,
-            value=auth.get_mock_jwt_cookie('user', lifetime_in_seconds=-1, access_token='token'),
+            value=get_mock_jwt_cookie('user', lifetime_in_seconds=-1, access_token='token'),
         )
         response = client.get(uri)
         assert response.status_code == HTTPStatus.UNAUTHORIZED
--- a/tests/test_api/test_api_spec.py
+++ b/tests/test_api/test_api_spec.py
@@ -46,8 +46,8 @@ def test_invalid_cookie(client):
 def test_invalid_bearer(client):
     for uri in ENDPOINTS:
         response = client.get(uri, headers={'Authorization': 'Bearer BAD TOKEN'})
-        assert 'Invalid authorization token provided' in response.json['detail']
         assert response.status_code == HTTPStatus.UNAUTHORIZED
+        assert 'Invalid authorization token provided' in response.json['detail']
 
 
 def test_expired_cookie(client):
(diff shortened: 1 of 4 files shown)


tests/test_api/test_auth.py
```diff
@@ -0,0 +1,50 @@
...
+    assert user == 'test-user'
+    assert token == 'test-token'
+
+
+@pytest.mark.network
+def test_decode_edl_bearer_token(monkeypatch, jwks_client):
+    def mock_decode(*args, **kwargs):
+        return {'uid': 'test-user'}
+
+    monkeypatch.setattr(auth.jwt, 'decode', mock_decode)
```
jtherrmann, line 22: It feels like we're mocking out a big chunk of the functionality by mocking `jwt.decode` in these two tests. I wonder how difficult it would be to come up with an actual token and cookie that could be decoded for these tests. Or would we run into an expiration date at some point?

tests/test_api/test_auth.py
```diff
@@ -0,0 +1,50 @@
...
+
+    user, token = auth.decode_edl_bearer_token('test-token', jwks_client)
+    assert user == 'test-user'
+    assert token == 'test-token'
```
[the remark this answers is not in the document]
jtherrmann: But this would be superseded by my comment regarding not mocking `jwt.decode`, below.

tests/test_api/test_api_spec.py
```diff
@@ -40,6 +39,14 @@ def test_invalid_cookie(client):
     for uri in ENDPOINTS:
         client.set_cookie(domain='localhost', key=AUTH_COOKIE, value='garbage I say!!! GARGBAGE!!!')
         response = client.get(uri)
+        assert 'Invalid authorization cookie provided' in response.json['detail']
+        assert response.status_code == HTTPStatus.UNAUTHORIZED
```
jtherrmann, line 43:
```suggestion
        assert response.status_code == HTTPStatus.UNAUTHORIZED
        assert 'Invalid authorization cookie provided' in response.json['detail']
```

tests/test_api/test_api_spec.py
```diff
@@ -40,6 +39,14 @@ def test_invalid_cookie(client):
...
+
+
+def test_invalid_bearer(client):
+    for uri in ENDPOINTS:
+        response = client.get(uri, headers={'Authorization': 'Bearer BAD TOKEN'})
+        assert 'Invalid authorization token provided' in response.json['detail']
         assert response.status_code == HTTPStatus.UNAUTHORIZED
```
jtherrmann, line 50:
```suggestion
        assert response.status_code == HTTPStatus.UNAUTHORIZED
        assert 'Invalid authorization token provided' in response.json['detail']
```

jtherrmann (commented):
I usually check the status code before the error message:

tests/test_api/test_auth.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@pytest.mark.network
+def test_get_jwks_client(jwks_client):
+    assert 'urs.earthdata.nasa.gov' in jwks_client.uri
+
+
+@pytest.mark.network
+def test_bad_bearer_token(jwks_client):
+    with pytest.raises(auth.InvalidTokenException, match='Invalid authorization token provided'):
```
jtherrmann, line 40:
```suggestion
    with pytest.raises(auth.InvalidTokenException, match=r'^Invalid authorization token provided'):
```

tests/test_api/test_auth.py
```diff
@@ -0,0 +1,51 @@
...
+        auth.decode_edl_bearer_token('bad token', jwks_client)
+
+
+def test_bad_asf_cookie(jwks_client, monkeypatch):
+    with pytest.raises(auth.InvalidTokenException, match='Invalid authorization cookie provided'):
```
jtherrmann, line 45:
```suggestion
    with pytest.raises(auth.InvalidTokenException, match=r'^Invalid authorization cookie provided'):
```

jtherrmann (commented):
I was thinking we might match against the full error message, but it looks like it would just be `Invalid authorization cookie provided: Not enough segments` (and similar for the token), which isn't very interesting. So I think let's just add the `^` to mark the start of the message:

tests/test_api/test_auth.py
```diff
@@ -0,0 +1,51 @@
...
+    assert 'urs.earthdata.nasa.gov' in jwks_client.uri
+
+
+@pytest.mark.network
+def test_bad_bearer_token(jwks_client):
+    with pytest.raises(auth.InvalidTokenException, match=r'^Invalid authorization token provided'):
+        auth.decode_edl_bearer_token('bad token', jwks_client)
+
+
+def test_bad_asf_cookie(jwks_client, monkeypatch):
```
jtherrmann, line 44:
```suggestion
def test_bad_asf_cookie(monkeypatch):
```

tests/test_api/test_auth.py
```diff
@@ -0,0 +1,51 @@
+from unittest.mock import patch
+
+import jwt
+import pytest
+
+from hyp3_api import auth
+from test_api import conftest
+
+
+def test_decode_asf_cookie(monkeypatch, jwks_client):
```
jtherrmann, line 10:
```suggestion
def test_decode_asf_cookie():
```

tests/test_api/test_auth.py
```diff
@@ -0,0 +1,51 @@
...
+    assert 'urs.earthdata.nasa.gov' in jwks_client.uri
+
+
+@pytest.mark.network
+def test_bad_bearer_token(jwks_client):
+    with pytest.raises(auth.InvalidTokenException, match=r'^Invalid authorization token provided'):
+        auth.decode_edl_bearer_token('bad token', jwks_client)
+
+
+def test_bad_asf_cookie(monkeypatch):
```
jtherrmann, line 44:
```suggestion
def test_bad_asf_cookie():
```

jtherrmann (approved):

PR #2826, opened 2025-06-26
Ensure OpenAPI Specification version is specified consistently

jhkennedy:
Since both places where we specify which version of the OpenAPI Specification we use are already Jinja templates, I've added an argument to `render.py` to specify the version and use that in the templates to ensure consistency.

I've also set the default version to v3.0.4, the current 3.0 series version.


jtherrmann (approved):
LGTM!

PR #2846, opened 2025-07-29
Use Graviton Instances for AUTORIFT_ITS_LIVE

AndrewPlayer3:


AndrewPlayer3:
As a result of this change, `cfn-lint` says a list is too long. This list is `TARGETS` in the following:

```
  Schedule:
    Type: AWS::Events::Rule
    Properties:
      ScheduleExpression: "rate(30 minutes)"
      Targets:
        - Arn: !GetAtt Lambda.Arn
          Id: lambda
        - Arn: !GetAtt AutoriftGravitonLambda.Arn
          Id:  AutoriftGravitonlambda
        - Arn: !GetAtt InsarIsceAriaLambda.Arn
          Id:  InsarIsceArialambda
        - Arn: !GetAtt AriaS1GunwLambda.Arn
          Id:  AriaS1Gunwlambda
        - Arn: !GetAtt SrgGslcLambda.Arn
          Id:  SrgGslclambda
        - Arn: !GetAtt SrgTimeSeriesLambda.Arn
          Id:  SrgTimeSerieslambda
```

It has a maximum length of 5: https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-events-rule.html#cfn-events-rule-targets

This means we can't have more than custom 4 compute environments for a given deployment, at a time.

apps/scale-cluster/scale-cluster-cf.yml.j2
```diff
@@ -177,11 +174,19 @@ Resources:
...
+          Id:  {{ name }}lambda
+
   {{ name }}EventPermission:
     Type: AWS::Lambda::Permission
     Properties:
       FunctionName: !GetAtt  {{ name }}Lambda.Arn
       Action: lambda:InvokeFunction
       Principal: events.amazonaws.com
-      SourceArn: !GetAtt Schedule.Arn
+      SourceArn: !GetAtt {{ name }}Schedule.Arn
```
AndrewPlayer3, line 191: This addresses the max list length error. As far as I can tell, it was probably supposed to be this the whole time. The previous method was using the scaling lambda for the default compute env for all of the compute envs.

jhkennedy (changes requested):
I didn't realize we were limited to 5 compute environments before! I like the scale cluster change to avoid this. 

however, I think we don't need to add a new compute environment for `AUTORIFT_ITS_LIVE.yml` as it's the only job in the its-live deployments:
https://github.com/ASFHyP3/hyp3/blob/develop/.github/workflows/deploy-enterprise.yml#L25-L27

So we can just change the default instance types and ami id for these deployments:
* hyp3-its-live
* hyp3-its-live-test


And for the AMI id, we should be able to use:
```
/aws/service/ecs/optimized-ami/amazon-linux-2023/arm64/recommended/image_id
```
which right now will get you:
* id: [ami-09f911ee9e8d7378c](https://us-west-2.console.aws.amazon.com/ec2/home?region=us-west-2#ImageDetails:imageId=ami-09f911ee9e8d7378c); name: `amzn2-ami-ecs-hvm-2.0.20250725-arm64-ebs`

```
$ aws ssm get-parameter --name /aws/service/ecs/optimized-ami/amazon-linux-2/arm64/recommended/image_id \ 
    --region us-west-2 --query "Parameter.Value" --output text
ami-09f911ee9e8d7378c
```

I'm content to leave autoRIFT running on the `r6id` instance family for EDC and ARIA deployments. 

So I'd keep the scale cluster changes, but drop the `job_spec/*` changes in favor of adjusting the default environment for ITS_LIVE.

We should check with @ASFHyP3/tools to see if they care about or have opinions on the scale-cluster changes.

asjohnston-asf (approved):

PR #2884, opened 2025-08-25
Expand SNS Publish permissions

jtherrmann:


apps/compute-cf.yml.j2
```diff
@@ -179,7 +179,7 @@ Resources:
             Resource: !Sub "arn:aws:s3:::${ContentBucket}/*"
           - Effect: Allow
             Action: sns:Publish
-            Resource: !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:*"
+            Resource: !Sub "arn:aws:sns:*"
```
asjohnston-asf, line 182: No need for a `!Sub` here anymore
```suggestion
            Resource: arn:aws:sns:*
```

asjohnston-asf (approved):

PR #2902, opened 2025-09-17
Allow for utilization for Sentinel-1c in JPL Deployments of Hyp3

cmarshak:


cmarshak:
@jhkennedy @asjohnston-asf

We are trying to use our AWS credits before end of FY. Hoping we can merge this ASAP. Thanks for your help!

PR #2905, opened 2025-09-18
Increase throughput

cmarshak:
This is a stand-alone PR just for increased throughput for JPL deployments - should not impact on-demand system.


cmarshak:
@jhkennedy @asjohnston-asf

We are trying to use our AWS credits before end of FY. Hoping we can merge this ASAP. Thanks for your help!

jhkennedy (approved):

PR #2917, opened 2025-09-26
Lower throughput for a19

cmarshak:
We have successfully utilized our AWS credits so returning throughput to nominal 4000 vCPUs.

### DIFF
--- a/.github/workflows/deploy-custom-prod.yml
+++ b/.github/workflows/deploy-custom-prod.yml
@@ -101,7 +101,6 @@ jobs:
             instance_types: m6id.xlarge,m6id.2xlarge,m6id.4xlarge,m6id.8xlarge,m6idn.xlarge,m6idn.2xlarge,m6idn.4xlarge,m6idn.8xlarge
             default_max_vcpus: 4000  # Max: 13000
             expanded_max_vcpus: 4000  # Max: 13000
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-custom-prod.yml
```diff
@@ -99,8 +99,8 @@ jobs:
               job_spec/ARIA_S1_COSEIS.yml
               job_spec/INSAR_ISCE.yml
             instance_types: m6id.xlarge,m6id.2xlarge,m6id.4xlarge,m6id.8xlarge,m6idn.xlarge,m6idn.2xlarge,m6idn.4xlarge,m6idn.8xlarge
-            default_max_vcpus: 13000  # Max: 13000
-            expanded_max_vcpus: 13000  # Max: 13000
+            default_max_vcpus: 4000  # Max: 13000 - recall this number is divided by 4 to get max concurrent jobs
```
jhkennedy, line 102:
```suggestion
            default_max_vcpus: 4000  # Max: 13000
            expanded_max_vcpus: 4000  # Max: 13000
```
the divide by 4 part isn't a constant -- it depends on job specific settings, primarily the instance family specified for the compute environment it uses as well the memory reservation for the job. All that to say, I don't think this is an appropriate place for this comment as it's subject to change and likely to become outdated.

PR #2946, opened 2025-10-23
Add VolcSARvatory multiburst job spec

mfangaritav:
This PR is to add a new job spec that uses multiburst processing avoiding the 15 bursts limit

### DIFF
--- a/.github/workflows/deploy-custom-test.yml
+++ b/.github/workflows/deploy-custom-test.yml
@@ -28,8 +28,8 @@ jobs:
               job_spec/ARIA_S1_COSEIS.yml
               job_spec/INSAR_ISCE.yml
             instance_types: m6id.xlarge,m6id.2xlarge,m6id.4xlarge,m6id.8xlarge,m6idn.xlarge,m6idn.2xlarge,m6idn.4xlarge,m6idn.8xlarge
-            default_max_vcpus: 640
-            expanded_max_vcpus: 640
+            default_max_vcpus: 6400
+            expanded_max_vcpus: 6400
             required_surplus: 0
             security_environment: JPL-public
             ami_id: /aws/service/ecs/optimized-ami/amazon-linux-2023/recommended/image_id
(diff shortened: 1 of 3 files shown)


job_spec/VOLCSARVATORY_MULTI_BURST.yml
```diff
@@ -0,0 +1,78 @@
+VOLCSARVATORY_MULTI_BURST:
```
jhkennedy:
```suggestion
INSAR_ISCE_MULTI_BURST:
```
I'd keep this name the same so that it will just be the multiburst job in the volcsarvatory deployment

.github/workflows/deploy-custom-test.yml
```diff
@@ -119,6 +119,7 @@ jobs:
             job_files: >-
               job_spec/INSAR_GAMMA.yml
               job_spec/INSAR_ISCE_BURST.yml
+              job_spec/VOLCSARVATORY_MULTI_BURST.yml
```
jhkennedy:
```suggestion
              job_spec/VOLCSARVATORY_MULTI_BURST.yml
```

Don't think we need two version of the multiburst job in this deployment.

job_spec/VOLCSARVATORY_MULTI_BURST.yml
```diff
@@ -0,0 +1,78 @@
...
+  required_parameters:
+    - reference
+    - secondary
+  parameters:
+    reference:
+      api_schema:
+        type: array
```
jhkennedy:
```suggestion
        type: array
        minItems: 1
        maxItems: 30
```
set_batch_overrides will fail if there are more than 30 items in this array, and that will happen when the job is executed, so we should add this so the API reject those kinds of jobs

https://github.com/ASFHyP3/hyp3/blob/develop/apps/set-batch-overrides/src/set_batch_overrides.py#L47-L73

PR #2959, opened 2025-11-03
drop unused services sandbox deploy workflow

jhkennedy:
The referenced deploy branch doesn't exist here in ASFHyP3 or in the ASFOpenSARlab fork, so this deploy workflow is unused as far as I can tell.


dgpalmieri (approved):
Makes sense to me, we don't have any plans to continue developing our HyP3 front end at this time

PR #2985, opened 2025-12-02
Add hyp3-edc-sandbox deployment

jtherrmann:
TODO:

- [x] Finish updating SSL cert section of README deployment guide
- [x] Confirm link to EDC article works after https://github.com/ASFHyP3/.github-private/pull/33 is merged
- [x] Delete `hyp3-jth-sandbox` CF stack, GH env, etc.
- [x] API endpoint
- [x] S3 endpoint
- [x] domain name
- [x] Run sample jobs, confirm succeed and can download files: https://hyp3-edc-sandbox.asf.alaska.edu/jobs?user_id=jtherrmann&name=first-sample

---

Only the OPERA_RTC_S1 job from the sample batch failed, not going to worry about that for now.


README.md
```diff
@@ -242,6 +233,8 @@ Go to AWS console -> Secrets Manager, then:
 
 #### Upload SSL cert
 
+TODO: finish updating this section; do we still reuse a master cert for all ASF-managed deployments?
+
```
asjohnston-asf: No, as of Dec all Tools-managed deployments use a deployment-specific cert provisioned through ACM.

SciDev deployments are still using older certs provisioned through inCommon, including the wildcard *.asf.alaska.edu cert, but per @jhkennedy they might transition to ACM in January.

jtherrmann: Thanks!

PR #3000, opened 2026-01-14
SlimSAR Time-domain Backprojection job spec file

ZachHoppinen:
This is PR:

- adds the job_spec file for running slimsar TDBP on files from s3 buckets
- adds the relevant job_spec to hyp3-slimsar-test deployment CI CD deployment matrix.
- adds correct container on ECR
- uses only GPU enabled nodes with it's own compute environment
- Runs the correct commands to the container including data_fps, nav_fp, dem_fp and an optional pols.

I most based this the other job specs. Especially RTC_Gamma.

I don't see any super relevant validators but how critical are those?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Added a `model_context_length` parameter to the `OPERA_DIST_S1` job specification.
 - Added the custom hyp3-slimsar-test deployment
-- Added new job_spec for slimsar time-domain backprojection processing
+- Added the new `SLIMSAR_TDBP` job_spec for slimsar time-domain backprojection processing
 
 ### Changed
 - Increased the maximum `stride_for_norm_param_estimation` to 32 from 16 for the `OPERA_DIST_S1` job specification to handle models with a larger input size.


job_spec/SLIMSAR_TDBP.yml
```diff
@@ -0,0 +1,70 @@
...
+          maxLength: 2
+          minLength: 2
+          pattern: "^(VV|VH|HV|HH)$"
+  validators: []
+  cost_profiles:
+    DEFAULT:
+      cost: 1.0
+  steps:
+    - name: ''
+      image: 242399506617.dkr.ecr.us-west-2.amazonaws.com/hyp3-slimsar
```
jhkennedy: I don't see `test` tagged image, so the `hyp3-slimsar-test` deployment won't find an image to pull, since it defaults to pulling `test`:
https://github.com/ASFHyP3/hyp3/pull/3000/changes#diff-1b786e2e386dfc278adcccdf6c86d74aa62c13e17abbefda42fcc94ada7d6b85R135

typically, test deployments default to pulling `test` tagged images and prod deployments default to pulling `latest` tagged images.

You can override the default in a job_spec with `image_tag:` as well.

job_spec/SLIMSAR_TDBP.yml
```diff
@@ -0,0 +1,70 @@
...
+    - nav_fp
+    - dem_fp
+  parameters:
+    data_fp:
+      api_schema:
+        type: string
+        example: s3://asf-bucket/slimsar/SAR_11072025_110035.sar
+        description: Path to uncompressed slimsar echo/raw file on s3 bucket
+        type: string
+        pattern: "^s3://[a-z0-9.-]+(/[^/]+)*/[^/]+\\.sar$"
```
jhkennedy:
```suggestion
        pattern: "^s3://[a-z0-9.-]+(/[^/]+)*/[^/]+\.sar$"
```

I think you have too many backslashes here:
https://regex101.com/r/ZYfYI7/1

job_spec/SLIMSAR_TDBP.yml
```diff
@@ -0,0 +1,70 @@
...
+      api_schema:
+        type: string
+        example: s3://asf-bucket/slimsar/NMZT23520001E_100_proj.dat
+        pattern:  "^s3://[a-z0-9.-]+(/[^/]+)*/[^/]+\\.dat$"
+        description: Path to post-processed nav .dat file on s3 bucket
+    dem_fp:
+      api_schema:
+        type: string
+        example: s3://asf-bucket/slimsar/repeat_sub.tif
+        pattern:  "^s3://[a-z0-9.-]+(/[^/]+)*/[^/]+\\.tif$"
```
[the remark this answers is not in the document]
jhkennedy: In the `--help` for the slimsar image, it has:
```
  --dem_fp DEM_FP       Path to DEM .tif/.dem file
```
do you wantt to support `*.dem` files in HyP3 too?

job_spec/SLIMSAR_TDBP.yml
```diff
@@ -0,0 +1,70 @@
...
+          - VV
+          - VH
+          - HV
+          - HH
+        default: [VV, VH, HV, HH]
+        items:
+          description: Polarization to process
+          maxLength: 2
+          minLength: 2
+          pattern: "^(VV|VH|HV|HH)$"
```
jhkennedy:
```suggestion
          type: string
          enum:
            - VV
            - VH
            - HV
            - HH
```
You can be explicit here with an enumerator as well. Either works, though.

job_spec/SLIMSAR_TDBP.yml
ZachHoppinen: Re uploaded with a `:test` appended.

PR #3008, opened 2026-01-28
Update SlimSAR instance types in compute_environments.yml

ZachHoppinen:
Removed additional instance types for SlimSAR configuration.


jhkennedy (approved):
Should be g2g; Just left a couple of take-it-or-leave-it comments.

PR #3027, opened 2026-02-12
Support for Arctic Trafficability stub-RTC plugin.

benbart:


job_spec/ARTRAFF_RTC.yml
```diff
@@ -0,0 +1,61 @@
+INSAR_ISCE_BURST:
```
jhkennedy:
```suggestion
ARTRAFF_RTC:
```
This is what the job name will actually be called in HyP3 -- the file name could be anything.

job_spec/ARTRAFF_RTC.yml
```diff
@@ -0,0 +1,59 @@
...
+        maxItems: 1
+        example:
+          - CAPELLA_C17_SM_SICD_HH_20251128093004_20251128093008.ntf
+        items:
+          description: Name of Capella granule
+          type: string
+          pattern: "^CAPELLA_.*.ntf"
+          minLength: 56
+          maxLength: 56
+          example: CAPELLA_C17_SM_SICD_HH_20251128093004_20251128093008.ntf
```
jhkennedy: This doesn't really need to be an array for just one item


```suggestion
  parameters:
    granule:
      api_schema:
        description: Name of Capella granule
        type: string
        pattern: "^CAPELLA_.*.ntf"
        minLength: 56
        maxLength: 56
        example: CAPELLA_C17_SM_SICD_HH_20251128093004_20251128093008.ntf
```

jhkennedy (approved):
Looks good! Going to merge this in so it can be added to TBD deployments and updated in conjunction with other HyP3 jobs specs.

PR #3029, opened 2026-02-18
Update dist s1 entrypoints

cmarshak:
Few questions:

1. I am not sure if my `command` will be problematic without a character in front of `run` for aws batch.
2. `dist_s1_data` option in `OPERA_DIST_S1_CONFIRMATION` is comma separated granules (could be up to 100 granules).

### DIFF
--- a/job_spec/OPERA_DIST_S1_CONFIRMATION.yml
+++ b/job_spec/OPERA_DIST_S1_CONFIRMATION.yml
@@ -1,15 +1,15 @@
-OPERA_DIST_S1:
+OPERA_DIST_S1_CONFIRMATION:
   required_parameters:
     - dist_s1_data 
   parameters:
     dist_s1_data:
       api_schema:
         type: array
-        minItems: 1
+        minItems: 2
         items:
           description: paths to an unconfirmed dist-s1 product
           type: string
-          example: "s3://hyp3-bucket/<job_id>/<dist_s1_product>"
+          example: "s3://hyp3-bucket/<job_id>/<dist_s1_product_directory>"
   cost_profiles:
     DEFAULT:
       cost: 1.0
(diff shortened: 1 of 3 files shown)


job_spec/OPERA_DIST_S1_CONFIRMATION.yml
```diff
@@ -0,0 +1,31 @@
...
+    - name: ''
+      image: ghcr.io/opera-adt/dist-s1
+      command:
+        - run_sequential_confirmation
+        - --dist_s1_data
+        - Ref::dist_s1_data
+        - --bucket
+        - '!Ref Bucket'
+        - --bucket_prefix
+        - Ref::job_id
```
jhkennedy: 🤔 I don't actually see container arguments for bucket and bucket prefix:

```shell
$ docker run --rm -it ghcr.io/opera-adt/dist-s1:test run_sequential_confirmation --help
Usage: python -m dist_s1 run_sequential_confirmation [OPTIONS]

  Run sequential confirmation of unconfirmed DIST-S1 products. Confirms
  products in order of oldest to newest.

Options:
  --dist_s1_data COMMA_LIST       Directory of OPERA products that are
                                  unconfirmed  [required]
  --dst_dist_product_parent TEXT  Path to parent directory for new DIST-S1
                                  product. If None, will be created in the
                                  current directory with the MGRS Tile ID as
                                  the name with first/last date.
  --no_day_limit INTEGER          No day limit in the confirmation logic -
                                  logic is constrained to this number.
  --exclude_consecutive_no_dist BOOLEAN
                                  Whether to exclude consecutive no
                                  disturbance.
  --percent_reset_thresh INTEGER  Percent reset threshold.
  --no_count_reset_thresh INTEGER
                                  No count reset threshold.
  --confidence_upper_lim INTEGER  Confidence upper limit.
  --confirmation_confidence_threshold FLOAT
                                  Confidence threshold.
  --metric_value_upper_lim FLOAT  Metric value upper limit.
  --help                          Show this message and exit.
```

Are they hidden?

PR #3071, opened 2026-04-10
Release v10.15.0

jhkennedy:
I've confirmed in the ITS_LIVE test deployment that the max_vcpus are being set correctly, and the compute envs now show the `r6id[n]` instances.


Alex-Lewandowski (approved):
LGTM!

Disclaimer: I am not certain that I would recognize an issue in job_spec/INSAR_ISCE_BURST_EC2.yml beyond line 34.

PR #3081, opened 2026-04-16
Fix S1C validation for SRG_TIME_SERIES and SRG_GSLC 

mfangaritav:
This PR is to fix a bug for `SRG_TIME_SERIES` and `SRG_GSLC` related to the S1C validation. The value for S1C is taken from https://sentiwiki.copernicus.eu/web/s1-products

### DIFF
--- a/apps/api/src/hyp3_api/validation.py
+++ b/apps/api/src/hyp3_api/validation.py
@@ -209,7 +209,7 @@ def check_same_relative_orbits(_, granule_metadata: list[dict]) -> None:
         name_split = granule['name'].split('_')
         absolute_orbit = name_split[7]
         # "Relationship between relative and absolute orbit numbers": https://sentiwiki.copernicus.eu/web/s1-products
-        offset = 73 if name_split[0] == 'S1A' else 27
+        offset = 73 if name_split[0] == 'S1A' else (172 if name_split[0] == 'S1C' else 27)
         relative_orbit = ((int(absolute_orbit) - offset) % 175) + 1
         if not previous_relative_orbit:
             previous_relative_orbit = relative_orbit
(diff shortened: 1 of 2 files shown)


apps/api/src/hyp3_api/validation.py
```diff
@@ -209,7 +209,7 @@ def check_same_relative_orbits(_, granule_metadata: list[dict]) -> None:
         name_split = granule['name'].split('_')
         absolute_orbit = name_split[7]
         # "Relationship between relative and absolute orbit numbers": https://sentiwiki.copernicus.eu/web/s1-products
-        offset = 73 if name_split[0] == 'S1A' else 27
+        offset = 73 if name_split[0] == 'S1A' else (172 if name_split[0] == 'S1C' else 27)
```
jhkennedy: Since we'll have 4 conditionals here in a bit, I think it might make sense to write this as:

```suggestion
        mission = name_split[0]
        if mission == 'S1A':
            offset = 73
        elif mission == 'S1B':
            offset = 27
        elif mission == 'S1C':
            offset = 172
        elif mission == 'S1D':
            offset = 42
        else:
            raise ValueError(f'Encountered unknown Sentinel-1 mission: {mission}')
```

The offset numbers look correct and match what's in ISCE3. I added S1D as well based on this ISCE2 PR:
https://github.com/isce-framework/isce2/pull/28123

The offset may change once S1D moves to its final orbit this summer, but this will allow us to process the commissioning data for now, so I think it's fine to add now and circle back if it needs to be updated.

jhkennedy (dismissed):
Looks good! I just added a suggestion which we could do now, or circle back to once S1D's orbit is final. Ether way works for me

jhkennedy (approved):

PR #3099, opened 2026-05-07
Add STREAM_FLOW.yml

kmarnoult:
Add STREAM_FLOW.yml to job_spec folder and update CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [10.16.5]
+
+### Added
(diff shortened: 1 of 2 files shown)


mfangaritav (dismissed):

jhkennedy (changes requested):
For this job type, we'll need to handle the input data some how. 

I think that would/could look like, either:
1. running the RTC and HydroSAR steps first, like in the [`WATER_MAP.yml`](https://github.com/ASFHyP3/hyp3/blob/develop/job_spec/WATER_MAP.yml) jobs
   * but we'd need to handle getting the approprite scenes for a node/reach, or take scene as input and decide the appropriate node/reach from there
2. gathering all the OPERA RTCs along the node/reach and merging them together, then running HydroSAR before running the streamflow code. 

The problem with (2), even though it's probably easier, is Fritz insists on doing this at 10m instead of the 30m provided by the OPERA RTCs, so we'll likely need to do (1).

PR #3128, opened 2026-06-10
Update pism-terra-execute job spec for upstream changes

jhkennedy:


kmarnoult (changes requested):
I think that text of line #7 in PISM_TERRA_EXECUTE.yml got rearranged.  It should probably be something line "S3 URI to run a script...".  Also, the word 'script' is missing the 't'.

jhkennedy:
> I think that the text of line 7 in PISM_TERRA_EXECUTE.yml got rearranged. It should probably be something like "S3 URI to run a script...". Also, the word 'script' is missing the 't'.

Ah good catch! Fixed.

kmarnoult (approved):

PR #3132, opened 2026-06-18
Add NISAR GUNW job spec to VolcSARvatory

mfangaritav:
This PR adds a new job spec to use `hyp3-isce3` for NISAR GUNW production.

### DIFF
--- a/job_spec/NISAR_GUNW.yml
+++ b/job_spec/NISAR_GUNW.yml
@@ -16,6 +16,8 @@ NISAR_GUNW:
     subset:
       api_schema:
         type: array
+        nullable: true
+        default: null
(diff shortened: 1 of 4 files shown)


job_spec/NISAR_GUNW.yml
```diff
@@ -0,0 +1,56 @@
...
+        - --bucket-prefix
+        - Ref::bucket_prefix
+        - --reference
+        - Ref::reference
+        - --secondary
+        - Ref::secondary
+        - --subset
+        - Ref::subset
+      timeout: 126000  # 35 hours
+      compute_environment: Default
```
ZachHoppinen, line 51: Will test with Default but may need to be more specific

ZachHoppinen (approved):

PR #3137, opened 2026-06-19
Change to granules for NISAR_GUNW

mfangaritav:
This PR changes the `reference` and `secondary` parameters in the  `NISAR_GUNW` job spec to `granules`


ZachHoppinen (approved):

jhkennedy:
... but why?

asjohnston-asf:
The api validator that checks if the granule exists didn't parse `reference` and `secondary` correct when they were strings; resulting in errors when trying to submit jobs.

PR #3175, opened 2026-09-02
Fix  /user endpoint lambda timeout from job name search

### LINKED ISSUE #1312: /user requests fail with "504 Server Error: Gateway Timeout" for users who have submitted many past jobs
**Jira:** https://asfdaac.atlassian.net/browse/TOOL-2041

The `/user` endpoint returns a list of all job `name`s ever submitted by the logged in user. For users that have submitted many (hundreds of thousands) of past jobs, the query for past job names can take longer than the 30 second API Gateway Integration timeout, resulting in a HTTP 504 Gateway Timeout error being returned to the user.

The troublesome query is `get_names_for_user()` invoked by the `/user` handler at https://github.com/ASFHyP3/hyp3/blob/develop/apps/api/src/hyp3_api/handlers.py#L84

Here is an example traceback generated for this error when calling `hyp3_sdk.hyp3.my_info()` as reported by @cirrusasf :
```
Traceback (most recent call last):
  File "/home/jiangzhu/apps/anaconda3/envs/its-live-operations/lib/python3.10/site-packages/hyp3_sdk/exceptions.py", line 29, in _raise_for_hyp3_status
    response.raise_for_status()
  File "/home/jiangzhu/apps/anaconda3/envs/its-live-operations/lib/python3.10/site-packages/requests/models.py", line 1022, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 504 Server Error: Gateway Timeout for url: https://hyp3-its-live.asf.alaska.edu/user

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jiangzhu/apps/anaconda3/envs/its-live-operations/lib/python3.10/site-packages/hyp3_sdk/hyp3.py", line 410, in my_info
    _raise_for_hyp3_status(response)
  File "/home/jiangzhu/apps/anaconda3/envs/its-live-operations/lib/python3.10/site-packages/hyp3_sdk/exceptions.py", line 33, in _raise_for_hyp3_status
    raise ServerError
hyp3_sdk.exceptions.ServerError
```

tylercchase:
The /user endpoint iterates through every job a user has. For users with _a lot_ of jobs this causes the request to hit a lambda timeout and error. This stops workflows in both the sdk and vertex interface.

This adds a time check to the function that gets jobs names that cuts off retrieval after 10 seconds pass and gives whatever job names it found during that time back to the user. The job names part of the user endpoint as far as I know is only used in vertex to fill in the project name selector option with some autofill options.

Fixes #1312

### DIFF
--- a/apps/api/src/hyp3_api/handlers.py
+++ b/apps/api/src/hyp3_api/handlers.py
@@ -14,6 +14,7 @@ from dynamo.exceptions import (
 from hyp3_api import util
 from hyp3_api.multi_burst_validation import MultiBurstValidationError
 from hyp3_api.validation import CmrError, ValidationError, validate_jobs
+import time
 
 
 def problem_format(status: int, message: str) -> Response:
@@ -128,9 +129,13 @@ def _user_response(user_record: dict) -> dict:
 
 def _get_names_for_user(user: str) -> list[str]:
     jobs, next_key = dynamo.jobs.query_jobs(user)
-    while next_key is not None:
-        new_jobs, next_key = dynamo.jobs.query_jobs(user, start_key=next_key)
-        jobs.extend(new_jobs)
+    start = time.time()
+    current_time = time.time()
+    if next_key:
+        while next_key is not None and current_time - start < 10:
+            new_jobs, next_key = dynamo.jobs.query_jobs(user, start_key=next_key)
+            jobs.extend(new_jobs)
+            current_time = time.time()
     names = {job['name'] for job in jobs if 'name' in job}
     return sorted(list(names))
 
--- a/apps/api/src/hyp3_api/handlers.py
+++ b/apps/api/src/hyp3_api/handlers.py
@@ -1,3 +1,4 @@
+import time
 from http.client import responses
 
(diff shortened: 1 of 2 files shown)


apps/api/src/hyp3_api/handlers.py
```diff
@@ -128,9 +129,13 @@ def _user_response(user_record: dict) -> dict:
 
 def _get_names_for_user(user: str) -> list[str]:
     jobs, next_key = dynamo.jobs.query_jobs(user)
-    while next_key is not None:
-        new_jobs, next_key = dynamo.jobs.query_jobs(user, start_key=next_key)
-        jobs.extend(new_jobs)
+    start = time.time()
+    current_time = time.time()
```
jhkennedy: ~~Shouldn't current time be in the while loop?~~ I can readz

Also, it might be necessary, but using `time.monotonic()` will protect against clock changes while running.

jhkennedy (approved):
This is great! I think the paging time limit makes sense.

PR #3179, opened 2026-09-15
adding GLACIER-MELT yml

jsingh2344:


job_spec/GLACIER_MELT.yml
```diff
@@ -0,0 +1,65 @@
...
+  cost_profiles:
+    DEFAULT:
+      cost: 1.0
+
+  validators: []
+
+  steps:
+    - name: ''
+      image: ghcr.io/jsingh2344/hyp3-glacier-melt
+      image_tag: 'test'
```
jhkennedy:
```suggestion
      # If left unspecified, HyP3 will automatically use the "deployment" level tag; typically `test` for test deployments and `latest` for production deployments. 
      # image_tag: v0.1.0   
```
If you remove this key entirely, HyP3 will automatically use the "deployment" level tag, which is typically `test` for test deployments and `latest` for production deployments.

